PR #2, opened 2016-11-17
Implement test requeueing

casperisfine:
### WAT DIZ?

All queue implementation get 2 new parameters:
  - `max_requeues` the maximum number of time a test can be requeued if unsuccessful.
  - `requeue_tolerance` the maximum percentage of tests runs that can be requeues, e.g. `0.01` for 1%. (The idea here is that something breaks pretty much all the tests of your test suite, you don't want to run your entire test suite multiple times).

### Limitations / Concerns

One thing that bother me, is that if a test is requeued, we need not to report (as in minitest reporter) it.

By that I mean that if a test fail but is requeued, it shouldn't appear in the `X tests, X assertions, X failures, X errors, X skips` summary and Minitest shouldn't know about it. Because if it succeed later, the summary and the exit code will be incorrect.

So it means our way to report failures to our exceptions tracker using a reporter won't work anymore.

cc @Shopify/pipeline thoughts ?


gcourtemanche:
LGTM 👍  Very nice improvement!
We need to fix summary before using it otherwise it's going to be very confusing.

lib/ci/queue/redis/worker.rb
```diff
@@ -28,18 +33,22 @@ def master?
...
           wait_for_master
-          while test = reserve
-            yield test
-            acknowledge(test)
+          until shutdown_required? || empty?
+            if test = reserve
+              yield test
+            else
+              sleep 0.05
+            end
```
wvanbergen: Why is this changed?

test/integration/redis_test.rb
```diff
@@ -23,24 +23,23 @@ def test_redis_reporter
...
 
       output = `ruby -Itest/fixtures test/fixtures/redis-runner.rb`.lines.map(&:strip).last
-      assert_equal '4 tests, 2 assertions, 1 failures, 1 errors, 1 skips', output
+      assert_equal '5 tests, 3 assertions, 1 failures, 1 errors, 1 skips', output
 
       io = StringIO.new
       summary.report(io: io)
       report = strip_heredoc <<-END
-        Ran 4 tests, 2 assertions, 1 failures, 1 errors, 1 skips in 0.00s (aggregated).
+        Ran 5 tests, 3 assertions, 1 failures, 1 errors, 1 skips in X.XXs (aggregated).
```
wvanbergen: Would it be possible to add the number of requeues in here?

lib/ci/queue/redis/worker.rb
wvanbergen: Nevermind, I think I get this now. The queue may be empty, but we have to wait for the retry queue to also be empty.

wvanbergen:
As for reporting, maybe we can do `class Requeued < Minitest::Skip`, wrap the failure inside this exception class and send that to the reporter?

casperisfine:
> We could still differentiate between skipped and requeued tests, but tools that are unaware will just treat it as a skipped test

🙏 that's the idea I was missing. Need to rework the code quite a bit to do that, but that's a neat idea.

kirs:
It would be also nice to have the requeued stats in the reporter:

```
Ran 5 tests, 3 assertions, 1 failures, 1 errors, 0 skips, 1 requeued
```

lib/minitest/queue.rb
```diff
@@ -4,6 +4,46 @@
...
+    def message
+      failure.message
+    end
+  end
+
+  module Requeueing
+    # Make requeues acts as skips for reporters not aware of the difference.
+    def skipped?
+      super || requeued?
+    end
```
wvanbergen: Is this method necessary? If a test is `requeued?` it should always be `skipped?` as well, right?

casperisfine: It's because of how the `skipped?` method is implemented in Minitest: https://github.com/seattlerb/minitest/blob/174a860e6c18162ebeb52c198dd45eb55eb8afcb/lib/minitest/test.rb#L246-L248

That code base is incredibly hard to extend.

wvanbergen: d'oh!

PR #4, opened 2016-11-23
Requeue tests with an offset

casperisfine:
### Why?

Since the worker is ready to pop a test right after it requeued a failing one, almost everytime it's the same worker that re-runs it.

It's not terrible, but it would be better if the test was re-ran by another worker, because if the test was failing due to some leaked state, another worker have more chances to make it pass.

Alternatively we could `lpush` the test back, but I'm not a fan since it mean all the requeued tests would be re-ran at the end of the build, so if you configured the queue to requeue tests more than once, it would have detrimental effect on performance.

Also you might wonder: why a static offset? 

Initially I wanted to randomize it based on the number of workers, but then realized we don't have that information ATM, and it seemed like another information to pass to the constructor for very little gain.

@Shopify/pipeline for review please.

### DIFF
--- a/lib/ci/queue/redis/worker.rb
+++ b/lib/ci/queue/redis/worker.rb
@@ -5,6 +5,11 @@ module CI
     module Redis
       ReservationError = Class.new(StandardError)
 
+      class << self
+        attr_accessor :requeue_offset
+      end
+      self.requeue_offset = 42
+
       class Worker < Base
         attr_reader :total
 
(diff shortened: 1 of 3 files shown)


lib/ci/queue/redis/worker.rb
```diff
@@ -100,11 +105,24 @@ def should_requeue?(test)
...
 
-        def requeue(test)
+        REQUEUE = %{
+          local queue_key = KEYS[1]
+          local test = ARGV[1]
+          local offset = ARGV[2]
+
+          local pivot = redis.call('lrange', queue_key, -1 - offset, 0 - offset)[1]
+          if pivot then
+            redis.call('linsert', queue_key, 'BEFORE', pivot, test)
```
wvanbergen: If I understand the Redis docs properly, this is an O(N) complex operation, right?.

wvanbergen:
Potential :bike::house: but what about an offset of 1? This would be a more straightforward implementation of the "next test will be consumed by same worker, so put it one after" assumption.

casperisfine:
> Potential 🚲🏠 but what about an offset of 1?

Yes, but the smaller the offset, the most likely the same worker picks it up back.

lib/ci/queue/redis/worker.rb
casperisfine: Yes it is. With N being `size(queue) - offset`.

I wish there was an index based insert, but there isn't :/

casperisfine: Actually wait, there is a faster way: http://stackoverflow.com/a/21695153

I'll change this.

wvanbergen:
Final 🚲 🏠 idea: can we wait with pushing it back on the queue until after we popped the next test to run off the queue? Probably too complex for this purpose, but that would give us the best guarantee.

lib/ci/queue/redis/worker.rb
casperisfine: Nevermind, it's actually similar complexity.

casperisfine:
> Final 🚲 🏠 idea: can we wait with pushing it back on the queue until after we popped the next test to run off the queue?

We could, but it's risky because it would mean losing atomicity (the requeued test would live solely in the worker memory for a while).

PR #5, opened 2016-11-23
Fix for missing variable

DrewMartin:
`msg` doesn't exist here. Was throwing errors for me locally.

@byroot


casperisfine:
:facepalm:

Thank you!

PR #7, opened 2016-12-01
Handle redis being down / unreachable

casperisfine:
If Redis is unreacheable, we assume that either:

  - It's just for that single worker, in this case other workers will pick up the slack.
  - It's actually down for everyone, the summary step will detect it.

@Shopify/pipeline for review please.

### DIFF
--- a/ci-queue.gemspec
+++ b/ci-queue.gemspec
@@ -23,7 +23,7 @@ Gem::Specification.new do |spec|
 
   spec.add_development_dependency 'bundler', '~> 1.13'
   spec.add_development_dependency 'rake', "~> 10.0"
-  spec.add_development_dependency 'minitest', '~> 5.0'
+  spec.add_development_dependency 'minitest', '~> 5.9.1'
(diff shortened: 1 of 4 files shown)


ci-queue.gemspec
```diff
@@ -23,7 +23,7 @@ Gem::Specification.new do |spec|
 
   spec.add_development_dependency 'bundler', '~> 1.13'
   spec.add_development_dependency 'rake', "~> 10.0"
-  spec.add_development_dependency 'minitest', '~> 5.0'
+  spec.add_development_dependency 'minitest', '~> 5.9.1'
```
casperisfine: Minitest released a `5.10.0` that breaks minitest-reporters, the fix is on the way upstream: https://github.com/kern/minitest-reporters/pull/220

casperisfine:
> If the first operation for the master fails (the one that sets the lock), it will not raise, because we expect a different one to become master, right?

Indeed. But if we acquired the master role, and then lose the connection before we can fill the queue, then we should let the exception bubble.

PR #10, opened 2017-02-13
Register workers on boot, surface on `Redis::Base`

DazWorrall:
@Shopify/pipeline for review. I wasn't 100% sure when is best to call `register`, but doing so after the tests are pushed means at least the environment is sane.

### DIFF
--- a/lib/ci/queue/redis/base.rb
+++ b/lib/ci/queue/redis/base.rb
@@ -42,6 +42,10 @@ module CI
           raise LostMaster, "The master worker is still `#{master_status}` after 10 seconds waiting."
         end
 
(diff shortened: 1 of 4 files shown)


lib/ci/queue/redis/base.rb
```diff
@@ -42,6 +42,10 @@ def wait_for_master(timeout: 10)
           raise LostMaster, "The master worker is still `#{master_status}` after 10 seconds waiting."
         end
 
+        def num_workers
```
casperisfine: I guess in Ruby `workers_count` would be more idiomatic /nitpick

PR #11, opened 2017-03-28
Write the test to the test order log file before it’s run, rather than after

wvanbergen:
This allows us to spot which tests was running in case of problems, like timeouts or segfaults.

### DIFF
--- a/lib/minitest/reporters/order_reporter.rb
+++ b/lib/minitest/reporters/order_reporter.rb
@@ -11,7 +11,8 @@ class Minitest::Reporters::OrderReporter < Minitest::Reporters::BaseReporter
     super
   end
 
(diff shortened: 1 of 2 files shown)


lib/minitest/reporters/order_reporter.rb
```diff
@@ -11,7 +11,8 @@ def start
     super
   end
 
-  def record(test)
+  def before_test(test)
+    super
     @file.puts("#{test.class}##{test.name}")
```
DazWorrall: Are these writes unbuffered?

casperisfine: They are buffered. But if I'm not mistaken `\n` tend to flush the buffer.

PR #16, opened 2017-06-14
Implement static and file and distributed queues in Python

casperisfine:
First parts of the Python implementation.

TODO:
  - [X] Build on travis
  - [x] README
  - [ ] ~~pytest plugin~~ (split to another PR)
    - [X] tests distribution
    - [ ] requeue support
    - [ ] Centralized error reports ?
  - [X] Package on Pypi (need to embed lua files)

cc @sol @sabidib


python/ciqueue/distributed.py
```diff
@@ -0,0 +1,148 @@
...
+    self._scripts = {}
+
+  def key(self, *args):
+    return ':'.join(['build', self.build_id] + map(str, args))
+
+  def wait_for_master(self, timeout=10):
+    if self.is_master:
+      return True
+
+    for i in range(timeout * 10 + 1):
```
sabidib: If you're not using `i` just put a `_` here. Is the idea to wait for 11*time.sleep(0.1) seconds?

python/ciqueue/distributed.py
```diff
@@ -0,0 +1,148 @@
...
+      master_status = self._master_status()
+      if master_status in ['ready', 'finished']:
+        return True
+      time.sleep(0.1)
```
sabidib: `1` instead of `0.1`  for a 10 second wait time.

python/ciqueue/distributed.py
```diff
@@ -0,0 +1,148 @@
...
+    return self.total - len(self)
+
+
+class Worker(Base):
+  def __init__(self, tests, worker_id, redis, build_id, timeout, max_requeues=0, requeue_tolerance=0):
+    super(Worker, self).__init__(redis=redis, build_id=build_id)
+    self.timeout = timeout
+    self.total = len(tests)
+    self.max_requeues = max_requeues
+    self.global_max_requeues = math.ceil(len(tests) * requeue_tolerance)
```
sabidib: What does `requeue_tolerance` represent?

python/ciqueue/pytest.py
```diff
@@ -0,0 +1,93 @@
...
+    'worker_id': args['worker'][0],
+    'build_id': args['build'][0],
+    'timeout': float(args.get('timeout', [0])[0]),
+    'max_requeues': int(args.get('max_requeues', [0])[0]),
+    'requeue_tolerance': float(args.get('requeue_tolerance', [0])[0]),
+  }
+
+
+def build_queue(queue_url, tests_index):
+  spec = urlparse(queue_url)
```
sabidib: What are the valid values that a --queue provides? Documenting them somewhere (docstring or argparse) might be useful.  They seem to be list://colon_seperated_strings, file://file_with_list_of_strings and redis://url ?

python/ciqueue/pytest.py
```diff
@@ -0,0 +1,93 @@
...
+  if spec.scheme == 'list':
+    return ciqueue.Static(spec.path.split(':'))
+  elif spec.scheme == 'file':
+    return ciqueue.File(spec.path)
+  elif spec.scheme == 'redis':
+    import redis
```
sabidib: Let's import this at the toplevel.

python/ciqueue/__init__.py
```diff
@@ -0,0 +1,3 @@
+from ciqueue.static import Static
```
casperisfine: I just didn't wan't to expose the internal file structure.

I'd agree for the `import distributed` though, since it has a dependency on `redis`, it's probably better not to import it by default

python/ciqueue/distributed.py
casperisfine: Nah it's correct. The idea is to check the status every 100ms up to `timeout`. So we sleep for 100ms (0.1) up to `timeout * 10` times.

python/ciqueue/distributed.py
casperisfine: It's for the requeue feature. The `pytest` plugin doesn't support it yet (and probably won't for the first version). 

The idea is that you can allow tests to be put back in the queue if they failed, so that if it's a leaky test it will have better chances to succeed.

This value sets a global % of tests that can be requeued. E.g. you have 200 tests, you set this to `0.1` to allow up to 20 requeues per builds.

redis/acknowledge.lua
```diff
@@ -1,8 +1,7 @@
 local zset_key = KEYS[1]
 local processed_key = KEYS[2]
 
-local worker_id = ARGV[1]
-local test = ARGV[2]
+local test = ARGV[1]
```
solackerman: why is this changing? And why is there lua (I'm new to lua)?

python/ciqueue/distributed.py
```diff
@@ -0,0 +1,165 @@
...
+    def requeue(self, test, offset=42):
+        return self._eval_script(
+            'requeue',
+            keys=[
+                self.key('processed'),
+                self.key('requeues-count'),
+                self.key('queue'),
+                self.key('running')],
+            args=[self.max_requeues, self.global_max_requeues, test, offset],
+        ) == 1
```
casperisfine: We could raise indeed, but booleans work quite well here:
  - True -> the test was requeued
  - False -> the test wasn't requeued (because the limit was reached already)

redis/acknowledge.lua
casperisfine: That's my bad, it was an unused variable and I noticed it while I was implementing this PR.

solackerman: Makes sense 👍 So how can I look at some code and say "this part would be better if it were written in lua"?

casperisfine: It's not about Lua vs Python. Most of what is done in Lua couldn't be done in Python at all.

Those lua scripts are executed inside Redis itself.

python/requirements.txt
```diff
@@ -0,0 +1,4 @@
+pytest==2.9.1
+setuptools==18.5
```
solackerman: shopify_python needs a more recent version of ci.
`setuptools==34.3`
ex. https://github.com/Shopify/starscream/blob/master/requirements/pip.txt#L2

Sorry, we should have mentioned that earlier.

solackerman:
Code looks great! 🚢 on 🍏 
My only request is for a readme with a brief explanation of how to use the api's.

casperisfine:
@solackerman @sabidib I've fixed the build (It uses `sudo: true` though and I'm not super happy with it).

I also added a README explaining how all this works.

And finally, I moved the pystest plugin effort in another PR: https://github.com/Shopify/ci-queue/pull/18 . Since you guys have some experience with https://github.com/Shopify/pytest-redis, you might have better luck that me at pulling it of.

Let me know what you think of all this.

python/README.md
```diff
@@ -0,0 +1,113 @@
+# ciqueue
```
solackerman: This readme is wickid!

.travis.yml
```diff
@@ -1,9 +1,18 @@
...
 rvm:
   - '2.3'
 before_install:
-  - cd ruby
   - gem install bundler -v 1.5.1
+  - cd python
+  - sudo pip install -U pip 
+  - sudo pip install setuptools==33.1.1
+  - pip --version
+  - sudo pip install -r requirements.txt
```
casperisfine: Yeah, I enqueued a couple dozen builds to get this working. There is likely a sudo free version that would work, but I don't know enough about all that stack to not spend days on it.

casperisfine:
> So just to be clear, we need #18 to go in as well before we can integrate this in our projects?

Yes. But I don't think I can pull off #18 alone. I'd be happy to give the necessary direction and support etc. But I'd appreciate if one of you could take over.

PR #18, opened 2017-06-28
Pytest plugin

casperisfine:
Work in progress, depends on https://github.com/Shopify/ci-queue/pull/16


.travis.yml
```diff
@@ -1,9 +1,18 @@
...
-  - cd ruby
   - gem install bundler -v 1.5.1
+  - cd python
+  - sudo pip install -U pip
+  - sudo pip install setuptools==33.1.1
+  - pip --version
+  - sudo pip install -r requirements.txt
+  - cd ../ruby
+
+after_script:
```
solackerman: @casperisfine when this is run in travis and a test fails, it doesn't fail the build. Do we know away around this issue (ie travis support for multiple projects in one repo)?

casperisfine: My bad. I tried very hard to got both test suites running at once, but couldn't find anything better. If I knew a better way to do this I would have used it.

solackerman: Ya, I couldn't find a quick solution either. 
How do you want to deal with this going forward? Should we split this up into separate projects, or should we move towards a build system that supports both?

casperisfine: The advantage of having both in the same repo is that we can easily share the lua scripts.

So it might be worth trying CircleCI.

Or we can split the 2 version and use a submodule or something like that for sharing the lua scripts.

casperisfine:
Also maybe it isn't failing because I use before / after ascript hooks. Maybe we could replace the script, and do both inside

solackerman:
@casperisfine are you OK with merging #16 to make reviewing this easier?

casperisfine:
Of course, gi ahead

python/ciqueue/pytest.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+class ItemList(object):
+
+    def __init__(self, index, queue):
+        self.index = index
+        self.queue = queue
+
+    def __len__(self):
+        # HACK: Prevent pytest from grabbing the next test
```
cfournie: Is this needed for `pytest` version 2? 3? Or both?

python/ciqueue/_pytest/outcomes.py
```diff
@@ -0,0 +1,48 @@
...
+
+class Failed(Exception):
+    "placeholder for runner.Failed which is not serializable"
+
+
+class UnserializableException(Exception):
+    "placeholder for any Exceptions that cannnot be serialized"
+
+
+SER = {runner.Skipped: Skipped,
```
cfournie: Can we expand these names? E.g. `SERIALIZE_TYPE`

python/ciqueue/pytest_report.py
```diff
@@ -20,6 +26,9 @@ def noop():
 
 @pytest.hookimpl(trylast=True)
 def pytest_collection_modifyitems(session, config, items):  # pylint: disable=unused-argument
+    # this function hooks into pytest's list of tests to run, converts all of them into
```
cfournie: Maybe make this a docstring (i.e. `"""`) and not just a comment?

solackerman:
@sabidib Can you :eyes: one last time before 🚢 ?

PR #27, opened 2017-08-24
retry failed tests in python

solackerman:
This is what it looks like on the worker agents:
image
In this example, 3 tests are retried, and marked as skipped.

### DIFF
--- a/python/ciqueue/pytest.py
+++ b/python/ciqueue/pytest.py
@@ -55,16 +55,28 @@ class ItemList(object):
     def __iter__(self):
         for test in self.queue:
             yield self.index[test]
-            # TODO: Find proper hook for acknowledge / requeue # pylint: disable=fixme
-            self.queue.acknowledge(test)
 
 
 class RedisReporter(object):
 
     def __init__(self, config, queue):
         self.config = config
+        self.queue = queue
         self.redis = queue.redis
         self.errors_key = queue.key('error-reports')
+        self.terminalreporter = config.pluginmanager.get_plugin('terminalreporter')
+        self.terminalwriter = config.get_terminal_writer()
+
+    def record(self, item):
+        # if the test passed, we remove it from the errors queue
+        # otherwise we add it
+        if hasattr(item, 'error_reports'):
+            self.redis.hset(
+                self.errors_key,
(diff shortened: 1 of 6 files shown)


python/ciqueue/pytest.py
```diff
@@ -80,13 +92,26 @@ def pytest_runtest_makereport(self, item, call):
...
+            # If the test was already acknowledged by another worker (we timed out)
+            # Then we only record it if it was successful.
+            elif self.queue.acknowledge(test_name) or not test_failed:
+                self.record(item)
+
+            # The test timed out and failed, mark it as skipped so that it doesn't
+            # fail the build
+            else:
+                outcomes.mark_as_skipped(call, item, self.terminalreporter.stats, "TIMED OUT")
+                self.terminalwriter.write(' TIMED OUT ', green=True)
```
cfournie: Why are we marking a test that isn't acknowledged (does this mean not running somewhere?) and failed as skipped?

casperisfine: So, if `acknowledge()` returned false, it mean another worker already acknowledged that test.

So based on the previous condition: `if self.queue.acknowledge(test_name) or not test_failed`

 - If the test was already acknowledged but is a success -> we record it because it either change nothing or flip a failure into a success.
  - If the test was already acknowledged but is a failure -> we drop it because it either change nothing or flip a success into a failure.

I know that if / else is a bit hard to grasp, that's why it's commented even in the ruby version. There might be a more readable way to express it but I haven't figured it out.

cfournie: So the condition where a test has failed for its 3rd time (i.e. `requeue` will return `False`) should return `True` on `acknowledge()` if no other worker has acknowledged it and it will get recorded on L108 as having failed, correct?

casperisfine: > a test has failed for its 3rd time

Assuming the requeue limit per test is configure at 2



> if no other worker has acknowledged it and it will get recorded on L108 as having failed, correct?

Correct.

cfournie: Thanks for the explanation 😄  Yeah that was a bit to unpack. I'm not sure how to make that clearer other than more comments.

Issue #30, opened 2017-09-07
don't push tests if there are failures during collection


solackerman:
eg.
https://buildkite.com/shopify/starscream/builds/80219#9bbff676-e0e4-4ad2-b121-45c8584f82de
In this instance, we know that there are errors during collection, but we push them anyway. The tests can't run on the workers, but by pushing them to the queue, an reporter is fooled into thinking that they'll eventually complete.

PR #31, opened 2017-09-19
small fixes

solackerman:
This PR fixes a bunch of little things with the python implementation:
 - it adds compression/decompression for the errors being pushed to redis. This is already a feature in the ruby implementation.
 - `key_item(item)` now returns pytest's builtin test identifier, instead of the something made up.
 - when reporting that a test will be retried, we will now output the message `WILL_RETRY` instead of `RETRYING`.
 - when an error occurs while collecting the tests, we no longer push the tests to ci anyways. This fixes https://github.com/Shopify/ci-queue/issues/30 where the summary agent waits indefinitely for the tests to complete.

### DIFF
--- a/python/ciqueue/_pytest/test_queue.py
+++ b/python/ciqueue/_pytest/test_queue.py
@@ -11,8 +11,7 @@ class InvalidRedisUrl(Exception):
 
 
 def key_item(item):
-    # TODO: discuss better identifier # pylint: disable=fixme
-    return item.location[0] + '@' + item.location[2]
+    return item.nodeid
 
 
 def parse_worker_args(query_string, tests_index):
--- a/python/ciqueue/pytest.py
+++ b/python/ciqueue/pytest.py
@@ -5,6 +5,7 @@ py.test -p ciqueue.pytest --queue redis://<host>:6379?worker=<worker_id>&build=<
 """
 from __future__ import absolute_import
 from __future__ import print_function
+import zlib
 from ciqueue._pytest import test_queue
 from ciqueue._pytest import outcomes
 import dill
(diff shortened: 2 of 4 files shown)


python/ciqueue/pytest.py
```diff
@@ -44,11 +45,6 @@ def __init__(self, index, queue):
         self.index = index
         self.queue = queue
 
-    def __len__(self):
-        # HACK: Prevent pytest from grabbing the next test
-        # TODO: Check if we could return a fake next test instead # pylint: disable=fixme
-        return 0
```
solackerman: We no longer need to hack the length to get around [this call](https://github.com/pytest-dev/pytest/blob/master/_pytest/main.py#L168) because that entire function is being overriden on line 114 below

python/ciqueue/pytest.py
```diff
@@ -114,12 +110,25 @@ def pytest_runtest_makereport(self, item, call):
                 self.terminalwriter.write(' TIMED OUT ', green=True)
 
 
-@pytest.hookimpl(trylast=True)
-def pytest_collection_modifyitems(session, config, items):
-    """This function hooks into pytest's list of tests to run, and replaces
-    those `items` with a redis test queue iterator."""
-    tests_index = ItemIndex(items)
+@pytest.hookimpl(tryfirst=True)
+def pytest_runtestloop(session):
```
solackerman: 90% copy-pasta'd from [here](https://github.com/pytest-dev/pytest/blob/master/_pytest/main.py#L158-L172) but with our ci queue inserted after we check for collection errors.

python/ciqueue/_pytest/test_queue.py
```diff
@@ -11,8 +11,7 @@ class InvalidRedisUrl(Exception):
 
 
 def key_item(item):
-    # TODO: discuss better identifier # pylint: disable=fixme
-    return item.location[0] + '@' + item.location[2]
+    return item.nodeid
```
solackerman: example key:
`integrations/pytest/test_all.py::TestHappy::()::test_method`

Issue #19, opened 2017-07-14
Separate populating the queue from working off the queue


wvanbergen:
I think I'd like to split up populating the queue, and working off the queue, so we can put a filter/order step in between. This allows us to remove tests we don't want to run, or order the list so the slowest ones get run first.

1. Retrieve list of tests. [Implementations: Minitest discovery, File, ...]
2. Annotate tests [e.g. Get statistics about every test case]
3. Filter/order based on annotations.
4. Populate to worker queue [Implementations: Memory, Redis]

Then, the existing worker code can work off this queue.

- For this to work, we probably need to wrap every test case in a class, rather than representing it as a string, e.g. `CI::Queue::Entry`. This class would hold all annotation.
- Annotations can come from the input list. This could be a more elaborate file format. For the minitest discovery, we could annotate test methods in the Ruby code.
- Annotations can also come from external sources as a separate step. E.g. retrieve the duration data for all tests in the list from an external database.

@DazWorrall @casperisfine what do you think?

casperisfine:
It's already split off, here's how we initialize the queue right now:

```
  Minitest.queue = CI::Queue::Redis.new(
    TestGlobs.all_tests(seed: ENV['BUILDKITE_COMMIT']),
    build_id: build_id,
    worker_id: ENV.fetch('BUILDKITE_PARALLEL_JOB'),
    redis: CIRedis.connection,
    **queue_config,
  )
```

All queue implementations just stupidly take a list of strings (test identifiers). If we want to filter out parts of the tests, it can and should be done beforehand.

The only downside is that for resiliency purposes, all workers can potentially be elected master, which mean they all compute the list of tests, but only one actually get to populate the shared queue with it.

So if we were to have a costly way to reduce the list it would be a bit wasteful.

> what do you think?

IMO if we want to filter out the tests, it should be the responsibility of a dedicated service which can hold state, record all tests runs results and refine it's heuristics based on that data.

e.g. `curl https://example.com/commits/abcdef092332432/tests.txt`.

PR #34, opened 2017-11-16
Decouple queue instantiation and population

casperisfine:
Fix: https://github.com/Shopify/ci-queue/issues/19

This allows to check if the queue was exhausted without having to boot the entire test environment, which should allow a late worker to exit faster.

This PR also change the queue interface. Queues now receive and return arbitrary objects instead of strings. This will make it easier to implement adapters for test frameworks others than Minitest.

### DIFF
--- a/ruby/lib/ci/queue/redis/base.rb
+++ b/ruby/lib/ci/queue/redis/base.rb
@@ -7,8 +7,8 @@ module CI
           @build_id = build_id
         end
 
-        def empty?
-          size == 0
+        def exhausted?
+          queue_initialized? && size == 0
         end
 
         def size
@@ -22,7 +22,7 @@ module CI
           redis.multi do
             redis.lrange(key('queue'), 0, -1)
             redis.zrange(key('running'), 0, -1)
-          end.flatten.reverse
+          end.flatten.reverse.map { |k| index.fetch(k) }
         end
 
         def progress
@@ -32,8 +32,7 @@ module CI
         def wait_for_master(timeout: 10)
           return true if master?
           (timeout * 10 + 1).to_i.times do
-            case master_status
-            when 'ready', 'finished'
+            if queue_initialized?
               return true
             else
               sleep 0.1
@@ -56,6 +55,13 @@ module CI
 
         def master_status
           redis.get(key('master-status'))
+        end
+
+        def queue_initialized?
+          @queue_initialized ||= begin
+            status = master_status
+            status == 'ready' || status == 'finished'
+          end
         end
 
         def eval_script(script, *args)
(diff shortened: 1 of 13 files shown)


ruby/lib/ci/queue/redis/base.rb
```diff
@@ -7,8 +7,8 @@ def initialize(redis:, build_id:)
           @build_id = build_id
         end
 
-        def empty?
-          size == 0
+        def exhausted?
```
casperisfine: I renamed `empty?` to `exhausted?` to more clearly reflect that the queue is empty because it was fully worked off, and not because it's not populated yet.

casperisfine:
Not sure why, but the python test suite is failing:

```
pkg_resources.DistributionNotFound: The 'pycodestyle>=2.3' distribution was not found and is required by autopep8
```

@solackerman any idea?

solackerman:
@casperisfine try bumping this version number?
https://github.com/Shopify/ci-queue/blob/master/python/setup.py#L45

casperisfine:
That fixed it thanks.

ruby/lib/ci/queue/redis/base.rb
```diff
@@ -58,6 +57,13 @@ def master_status
           redis.get(key('master-status'))
         end
 
+        def master_elected?
+          @master_elected ||= begin
+            status = master_status
+            status == 'ready' || status == 'finished'
```
DazWorrall: Do we consider the master elected after the queue is populated (state == ready) or when a worker gets the lock to begin pushing the queue (state == setup)? Just want to make sure `master_elected?` is what you intend.

casperisfine: Technically, the master is indeed already elected when the status is set to `setup`.

However at that point the queue isn't filled, so yeah maybe that method should be along the lines of`master_elected_and_queue_initialized?` but it's seem a bit verbose.

WDYT?

wvanbergen: maybe just `queue_initialized?`.That implies that a master has been elected as well.

casperisfine: Makes sense.

ruby/test/support/shared_queue_assertions.rb
```diff
@@ -1,12 +1,36 @@
 module SharedQueueAssertions
+  class TestCase
```
wvanbergen: Does it make sense to have a test case class in the actual library, rather than just for test purposes?

casperisfine: It's a fake class. Since we're about to support multiple test frameworks, the queue implementations don't know what kind of objects they contain.

However we could expect an interface, and the different implementations would have to wrap whatever test case objects they have in some kind of delegate object to translate to and from string.

PR #35, opened 2017-11-16
Fix Python test suite

casperisfine:
As suggested by https://github.com/Shopify/ci-queue/pull/34#issuecomment-344943270


casperisfine:
cc @solackerman thanks for the tip!

PR #41, opened 2017-11-22
Document minitest-queue usage

casperisfine:
Pretty much all is in the title. The documentation might still be a little light but it's better than nothing I guess...

I think after this PR we can cut a new release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,31 +12,30 @@ There are algorithms available to balance perfectly your workers, but in practic
 
 Another advantage is that if you lose workers along the way, using a queue the other workers can pick up the job, making you resilient to failures.
 
-## Installation
+## How does it work?
 
-Add this line to your application's Gemfile:
+Each workers first participate in a leader election, the elected leader will then populate the queue with the list of tests to run.
+Once the queue is initialized, all workers including the leader will reserve tests and work them on until the queue is empty.
 
-```ruby
-gem 'ci-queue'
-```
+If a worker were to die, its reserved work load will be put back into the queue to be picked up by it's surviving siblings.
 
-And then execute:
+Additionally, a separate process can be started to centralize the error reporting, that process will wait for the queue to be empty before exiting.
 
-    $ bundle
+## What are requeues?
 
-Or install it yourself as:
+When working on big test suites, it's not uncommon for a few tests to fail intermittently, either because they are inherently flaky,
+or because they are sensible to other tests modifying some global state (leaky tests).
 
-    $ gem install ci-queue
+In this context, it is useful to have mitigation measures so that these intermittent failures don't cause unrelated builds to fail until the root cause is addressed.
 
-## Usage
+This is why `ci-queue` optionally allows to put failed tests back into the queue to retry them on another worker, to ensure the failure is consistent.
 
-TODO: Write usage instructions here
+## Installation and usage
 
-## Development
+Two implementations are provided, please refer to the respective documentations:
 
-After checking out the repo, run `bin/setup` to install dependencies. Then, run `rake test` to run the tests. You can also run `bin/console` for an interactive prompt that will allow you to experiment.
-
-To install this gem onto your local machine, run `bundle exec rake install`. To release a new version, update the version number in `version.rb`, and then run `bundle exec rake release`, which will create a git tag for the version, push git commits and tags, and push the `.gem` file to [rubygems.org](https://rubygems.org).
+  - [Python](python/)
+  - [Ruby](ruby/)
 
 ## Contributing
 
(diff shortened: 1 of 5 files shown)


ruby/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+Or install it yourself as:
+
+    $ gem install ci-queue
+
+## Usage
+
+### Supported CI providers
+
+`ci-queue` automatically infer most of it's configuration if ran on one of the following CI providers:
```
wvanbergen: Small `infers` & `its configuration`

ruby/README.md
```diff
@@ -0,0 +1,48 @@
...
+  - Travis
+
+If you are using another CI system, please refer to the command usage message.
+
+### Minitest
+
+Assuming you use one of the supported CI providers, the command can be as simple as:
+
+```bash
+minitest-queue --url redis://example.com run -Itest test/**/*_test.rb
```
wvanbergen: I find the meaning of `--url` a bit vague. Maybe instead of using `--url` we should use `--queue` ? Not sure.

ruby/lib/ci/queue/configuration.rb
```diff
@@ -14,7 +14,7 @@ def from_env(env)
       end
 
       def initialize(
-        timeout: 10, build_id: nil, worker_id: nil, max_requeues: 0, requeue_tolerance: 0,
+        timeout: 30, build_id: nil, worker_id: nil, max_requeues: 0, requeue_tolerance: 0,
```
wvanbergen: Why this change?

ruby/README.md
wvanbergen: Or, `--orchestrator` or `--coordinator`

ruby/lib/ci/queue/configuration.rb
casperisfine: Because I figured 10 seconds is very low, and people trying out this gem might be surprised by the behavior. 30 seemed fair.

ruby/README.md
casperisfine: Fixed.

ruby/README.md
casperisfine: Changed it for `--queue`

ruby/lib/minitest/queue/runner.rb
```diff
@@ -76,76 +80,141 @@ def report_command
...
+          EOS
+          opts.separator ""
+          opts.on('--worker WORKER_ID', *help) do |worker_id|
             queue_config.worker_id = worker_id
           end
 
+          help = split_heredoc(<<-EOS)
+            Defines how many time a single test can be requeued.
+            Defaults to 0.
+          EOS
```
wvanbergen: Not sure about the minimum Ruby version you had in mind, but if you'd use `<<~EOS` there's no need for stripping the string, e.g. `opts.on('--requeue-tolerance RATIO', *help.lines)` without `split_heredoc`.

README.md
```diff
@@ -12,37 +12,36 @@ There are algorithms available to balance perfectly your workers, but in practic
 
 Another advantage is that if you lose workers along the way, using a queue the other workers can pick up the job, making you resilient to failures.
 
-## Installation
+## How does it work?
 
-Add this line to your application's Gemfile:
+Every workers first participate in a leader election, the elected leader will then populate the queue with the list of tests to run.
```
DazWorrall: Suggest `Every workers first participate` -> `Each worker first participates`

README.md
```diff
@@ -12,37 +12,36 @@ There are algorithms available to balance perfectly your workers, but in practic
...
+Once the queue is initialized, all workers including the leader will reserve tests and work them on until the queue is empty.
 
-```ruby
-gem 'ci-queue'
-```
+If a worker were to die, it's reserved work load will be put back into the queue to be picked up by it's surviving siblings.
```
DazWorrall: Grammar suggestion (twice in this sentence): `it's` -> `its`

ruby/lib/minitest/queue/runner.rb
casperisfine: Lines have to be passed as an array of strings, otherwise the formatting is messed up. So it doesn't change much.

Also it would have been two small of an improvement for requiring 2.3. That's why I didn't use `<<~`

README.md
```diff
@@ -12,37 +12,36 @@ There are algorithms available to balance perfectly your workers, but in practic
...
 
-And then execute:
+Additionally, a separate process can be started to centralize the error reporting, that process will wait for the queue to be empty before exiting.
 
-    $ bundle
+## What are requeues?
 
-Or install it yourself as:
+When working on big test suites, it's not uncommon for a few tests to fail intermittently, either because they are inherently flaky,
+or either because they are sensible to other tests modifying some global state (leaky tests).
```
DazWorrall: `or either because they are sensible to` -> `or because they are sensitive to`

casperisfine:
Thanks @DazWorrall.

casperisfine:
I'll cut a release in a bit.

PR #43, opened 2017-11-23
Implement minitest-queue bisect

casperisfine:
This PR extract and properly package the `bisect-leak` script we've been using at Shopify.

Usage: `minitest-queue --queue path/to/test_order.log --failing-test 'Test#identifier' bisect -Itest path/to/*_test.rb`.

See `minitest_bisect_test.rb` for more detailed usage, and output.


ruby/lib/minitest/queue.rb
```diff
@@ -58,7 +58,8 @@ def queue=(queue)
 
     def queue_reporters=(reporters)
       @queue_reporters ||= []
-      Reporters.reporters = ((Reporters.reporters || []) - @queue_reporters) + reporters
+      Reporters.use!(((Reporters.reporters || []) - @queue_reporters) + reporters)
+      Minitest.backtrace_filter.add_filter(%r{exe/minitest-queue|lib/ci/queue/})
```
wvanbergen: This may only work in Rails environments. The minitest backtrace_filter object is not as advanced as the Rails one. Not sure if that's a problem.

casperisfine: No it just need `minitest-reporters` (which is a `ci-queue` dependency). The integration test goes through this and it works just fine.

PR #44, opened 2017-11-28
Initial RSpec integration

casperisfine:
As stated in the README, this is missing centralized reporting as well as requeueing. But those 2 are somewhat optional, and tricky to implement, so I'd rather work on them in followup PRs.

I don't except a throughout review, since most of the code is patching RSpec internals and that would take a while to grasp. But a quick sanity check is more than welcome.

### DIFF
--- a/ruby/lib/rspec/queue.rb
+++ b/ruby/lib/rspec/queue.rb
@@ -1,0 +1,202 @@
+require 'rspec/core'
+require 'ci/queue'
+
+module RSpec
+  module Queue
+    class << self
+      def config
+        @config ||= CI::Queue::Configuration.from_env(ENV)
+      end
+    end
+
+    module ConfigurationExtension
+      private
+
+      def command
+        'rspec' # trick rspec into enabling it's default behavior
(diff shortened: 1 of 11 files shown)


ruby/lib/rspec/queue.rb
```diff
@@ -0,0 +1,204 @@
...
+        example_group.set_ivars(instance, example_group.before_context_ivars)
+        succeeded = example.run(instance, reporter)
+        if !succeeded && reporter.fail_fast_limit_met?
+          RSpec.world.wants_to_quit = true
+        end
+        succeeded
+      end
+    end
+
+    URL = 'list:.%2Ftest%2Ffixtures%2Fdummy_spec.rb%5B1%3A1%5D:.%2Ftest%2Ffixtures%2Fdummy_spec.rb%5B1%3A2%5D'
```
DazWorrall: What is this used for?

casperisfine: That's some early test code left over -_-.

PR #45, opened 2017-11-30
Handle RSpec's ExampleGroup descendants (and fix shuffling)

casperisfine:
It's a bit contrived IMO, but turns out the `example_groups` list is not enough, each of them have "descendants" (nested describes), that you have to iterate over as well.

I also realized I forgot to implement test shuffling for RSpec....

### DIFF
--- a/ruby/lib/rspec/queue.rb
+++ b/ruby/lib/rspec/queue.rb
@@ -149,7 +149,7 @@ module RSpec
       end
 
       def run_specs(example_groups)
-        examples = example_groups.flat_map do |example_group|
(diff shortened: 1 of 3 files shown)


ruby/lib/rspec/queue.rb
```diff
@@ -174,6 +174,12 @@ def run_specs(example_groups)
 
       private
 
+      def shuffle(examples)
+        return tests unless RSpec::Queue.config.seed
```
DazWorrall: Should you return `examples` rather than `tests` here?

casperisfine: 🤦‍♂️

PR #46, opened 2017-12-01
Built-in SnappyPack codec

casperisfine:
We've seen a 78% reduction is memory usage when we enabled this on out application, so might as well provide it automatically in the gem.

However `snappy` and `msgpack` both being native gems, I'd rather not force the dependency. The optimization will be enabled automatically if the project uses those already. Otherwise it will fallback to `Marshal`.


casperisfine:
@wvanbergen done!

PR #47, opened 2017-12-12
Python3 compatibility.

erikwright:
@solackerman FYI for now. There are still two tests that fail, and I don't understand why. They seem to run the tests, but tests that are supposed to fail pass, so the result message doesn't contain the expected output. Maybe you will have an idea if you have a better understanding of how it works?

The main changes for Python 3 were:
1. Strings are unicode by default, so the byte strings returned from Redis need to be decoded. I tried setting automatic decoding using UTF-8 but that doesn't work if you store binary data (which you do, for the pickled test results).
2. The `urlparse` module was refactored. I added `future` which offers a compatibility layer.
3. `xrange` is renamed to `range`. `future` adds a compatibility layer that does the correct thing in Py2/Py3 (because in the situation where you use it, you have a potentially large range).

I included a `tox` file. You can run in Python 3 using `tox -e py36 -- tests/`.

---
This change is [Reviewable](https://reviewable.io/reviews/shopify/ci-queue/47)


.travis.yml
```diff
@@ -14,4 +20,4 @@ before_install:
   - cd ../ruby
 script:
   - cd ../ruby && bundle exec rake
-  - cd ../python && PYTHONPATH='.' make test
+  - cd ../python && PYTHONPATH='.' sudo make test
```
erikwright: Why is the `sudo` required?

python/ciqueue/distributed.py
```diff
@@ -70,14 +72,14 @@ def poll():
...
-                    yield test
+                    yield test.decode()
                 else:
                     time.sleep(0.05)
 
         try:
             self.wait_for_master()
             for i in poll():
-                yield i
+                yield i if isinstance(i, str) else i.decode()
```
erikwright: I suspect that I added this and it isn't required. From inspection, it seems `i` will always be decoded on line 75.

python/setup.cfg
```diff
@@ -1,3 +1,11 @@
+[tox]
+envlist = py27,py36
+
+[testenv]
+commands = pytest {posargs:-vv}
+passenv = REDIS_HOST
+deps = ipdb
```
sabidib: Why is ipdb required? Is this left over from debugging?

python/setup.py
```diff
@@ -15,11 +15,14 @@ def get_lua_scripts():
 
     paths = []
 
+    if not os.path.exists(SCRIPTS_PATH):
```
sabidib: Is this a bug fix? I might be missing how this is related to the python3 compat changes.

python/ciqueue/distributed.py
```diff
@@ -38,7 +40,7 @@ def wait_for_master(self, timeout=10):
             "` after {} seconds waiting.".format(timeout))
 
     def _master_status(self):
-        return self.redis.get(self.key('master-status'))
+        return self.redis.get(self.key('master-status')).decode()
```
sabidib: Hm, do we really have to call `.decode` anytime we read from redis? Is there a setting that can force redis to just always return a string?

.travis.yml
solackerman: tox creates a .tox folder and populates it with stuff (virtualenvs etc), and it needs sudo privileges to write there.

python/ciqueue/distributed.py
solackerman: I'll check.

python/ciqueue/distributed.py
solackerman: :+1: K, I'll try it out.

python/setup.py
erikwright: This packaging code was not previously called in CI, and it failed. Now that we use `tox`, we actually run `setup.py` in CI. So it's actually increased code coverage.

python/setup.cfg
erikwright: Yes, but ultimately it is needed anytime you want to debug a test, so I argue it makes sense to leave it. I'm open-minded. We don't have it in Crucible's CI tox env, so maybe we don't need it here.

python/ciqueue/distributed.py
erikwright: There is such a setting, but that causes failures if any of your values are _not_ unicode strings. CIQueue, as it happens, stores pickled Python objects as binary data.

python/setup.cfg
solackerman: I'll remove it, we'll see :)

python/setup.py
solackerman: I think it's a bug fix

.travis.yml
erikwright: Weird. Even to write in the source folder? Is there some other folder that one _is_ expected to write to during the build process?

Anyway, just curious. I can't think of any reason to care about it, it just seems bizarre.

solackerman: I'm not sure, but this would be a good time to mention that we're "off the travis rails".
Travis only supports one language per project in its test suites where everything is included. We have ruby and python libraries to test, so we're doing a lot more bash-ing and using it more like buildkite then a standard travis setup.

Issue #48, opened 2017-12-12
Starscream shows successfully retried tests as errors


cfournie:
On [otherwise good builds](https://buildkite.com/shopify/starscream/builds/93358), retried tests show up as failures in Starscream.

![screen shot 2017-12-12 at 12 27 05 pm](https://user-images.githubusercontent.com/49264/33898966-fdcaec4a-df37-11e7-93d8-439166b60b3b.png)

Despite having been retried successfully.

<details>

Node 2:
```
tests/starscream/loaders/test_sqlalchemy_loaders.py::TestRedshiftLoaderIntegration::test_load_from_scratch_incremental[parquet] FAILED WILL_RETRY 
tests/starscream/loaders/test_sqlalchemy_loaders.py::TestRedshiftLoaderIntegration::test_load_from_scratch_incremental[parquet] SKIPPED
```

Node 3:
```
tests/starscream/loaders/test_sqlalchemy_loaders.py::TestRedshiftLoaderIntegration::test_load_from_scratch_incremental[json] PASSED
```
</details>

cc/ @Shopify/dev-acceleration

wvanbergen:
How are JUnitXML files generated in your stack?

cfournie:
[`py.test` creates it](https://github.com/pytest-dev/pytest/blob/5631a8629696ec649b59feae99f7e7e057579c9b/_pytest/junitxml.py#L412-L434) for us after it's finished running our tests ([which we call in our test runner](https://github.com/Shopify/starscream/blob/ecffc5a7cff0386f96f6fd7944234c4db5940b56/.buildkite/runners/test#L32)).

wvanbergen:
In Ruby's MiniTest, we had to override some behaviour so it treats requeued/retried tests as "skipped" to make all the tooling that is unaware of retries work out of the box. My python knowledge and skills are not enough to know whether that is feasible in your case. What do you think?

cfournie:
We mark as skipped and requeue:

https://github.com/Shopify/ci-queue/blob/7a601e0402e0966000a317993d5f483175089217/python/ciqueue/pytest.py#L95-L99

But I guess not well enough because we're still writing out a failure :|

solackerman:
We modify the stats to look like the test never happened, but we don't remove the test itself.
https://github.com/Shopify/ci-queue/blob/master/python/ciqueue/_pytest/outcomes.py#L95-L97

jules2689:
[...]

PR #49, opened 2017-12-22
roll back junit trackers on skip, and add tests

solackerman:
fixes https://github.com/Shopify/ci-queue/issues/48

### DIFF
--- a/python/ciqueue/pytest.py
+++ b/python/ciqueue/pytest.py
@@ -62,6 +62,7 @@ class RedisReporter(object):
         self.errors_key = queue.key('error-reports')
         self.terminalreporter = config.pluginmanager.get_plugin('terminalreporter')
         self.terminalwriter = config.get_terminal_writer()
+        self.logxml = config._xml if hasattr(config, '_xml') else None  # pylint: disable=protected-access
 
     def record(self, item):
         # if the test passed, we remove it from the errors queue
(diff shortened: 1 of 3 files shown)


solackerman:
ping!

python/ciqueue/pytest.py
```diff
@@ -74,6 +75,43 @@ def record(self, item):
...
+        # the call is converted to a skip
+        call.excinfo = outcomes.skipped_excinfo(item, msg)
+
+        # clear out the stats like the test never happened
+        for key in ('passed', 'error', 'failed'):
+            clear_out_stats(key)
+
+        # rollback the testsfailed number like it never happened
+        item.session.testsfailed -= len([v for k, v in item.error_reports.items()
+                                         if not issubclass(v['excinfo'].type, outcomes.Skipped) and k != 'teardown'])
```
cfournie: Are failed teardown's counted elsewhere?

solackerman: We mark the test as skipped after setup and test, but before reporting on teardown. What we're doing here is rolling back setup and test, and replacing teardown's result with skipped. So the reason we don't have to rollback teardown is because we haven't reported on it yet. At the top of the function, we assert that this is only called on teardown.

solackerman:
We can can update the package and re-enable it.

PR #50, opened 2017-12-24
support up to pytest 3.3.1

solackerman:
This PR enables ciqueue to work with versions of pytest up until the latest (3.3.1)

### DIFF
--- a/python/ciqueue/_pytest/outcomes.py
+++ b/python/ciqueue/_pytest/outcomes.py
@@ -14,7 +14,10 @@ when reading off the queue. These operations are performed by
 
 from __future__ import absolute_import
 import dill
-from _pytest import runner
+try:
+    from _pytest import outcomes
+except ImportError:
+    from _pytest import runner as outcomes
 from _pytest._code import code
 from tblib import pickling_support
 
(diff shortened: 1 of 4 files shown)


python/ciqueue/_pytest/outcomes.py
```diff
@@ -14,36 +14,39 @@
 
 from __future__ import absolute_import
 import dill
-from _pytest import runner
+try:
+    from _pytest import outcomes
+except ImportError:
+    from _pytest import runner as outcomes
```
solackerman: This is the big change that we're now supporting. In pytest 3.2, they moved the `Failed` and `Skipped` classes from the `runner` module to `outcomes`.

python/ciqueue/pytest.py
```diff
@@ -61,8 +62,22 @@ def __init__(self, config, queue):
         self.redis = queue.redis
         self.errors_key = queue.key('error-reports')
         self.terminalreporter = config.pluginmanager.get_plugin('terminalreporter')
+        if hasattr(self.terminalreporter, '_get_progress_information_message'):
+            self.__replace_progress_message()
         self.terminalwriter = config.get_terminal_writer()
 
+    def __replace_progress_message(self):
```
solackerman: This is "gravy". In pytest 3.3 they added progress reporting to the terminal:
image
This change ensures that those numbers continue to represent the progress of the entire test suite, as opposed to the % of tests run on this agent relative to the total.

python/tests/test_pytest.py
```diff
@@ -5,7 +5,8 @@
 
 def expected_messages(output):
     assert '= 4 failed, 2 passed, 1 skipped, 1 xpassed, 6 error in' in output, output
-    assert 'integrations/pytest/test_all.py:27: skipping test message' in output, output
+    assert ('integrations/pytest/test_all.py:27: skipping test message' in output
+            or 'integrations/pytest/test_all.py:28: skipping test message' in output), output
```
solackerman: This change is because of a bug fix in pytest 3.2. Before then, the skip line was recorded as 1 line above the `pytest.skip` call. So the test's skip message will vary depending on the version of pytest it's running on.

python/ciqueue/pytest.py
cfournie: Is this output shown on the individual agents? Isn't it a little confusing to show overall progress on individual agents?

solackerman: This output is shown on the individual agents. 
Currently, the summary agent doesn't show progress. It waits until all the tests have completed before reporting on anything, and it would be a large change to alter that behaviour.

cfournie: Can you user test this? If there's confusion over what this number means we should omit it entirely. I'm worried that it will be confused as per-agent percentage and no one would guess that it represents overall build progress.

solackerman: I'm going to disable progress to unblock the crucible team and loop back in another PR (or not).

PR #52, opened 2018-01-08
Refactor travis test suite

casperisfine:
Context: Minitest 5.11 was released early this year, and broke the monkey patch we use in ci-queue to implement requeues. So in order to support both versions of minitest we'll need to test both on CI.

Because of this I decided to refactor the `travis.yml` a bit to makes testing 2 versions of minitest easier.

### DIFF
--- a/bin/before-install
+++ b/bin/before-install
@@ -1,0 +1,22 @@
+#!/usr/bin/env ruby
+
+def run(*args)
+  puts "$ #{args.join(' ')}"
+  unless system(*args)
+    exit $?.exitstatus
+  end
+end
+
+case ENV['SUITE']
+when 'python'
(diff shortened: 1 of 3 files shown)


bin/before-install
```diff
@@ -0,0 +1,23 @@
...
+    run('sudo', 'pip', 'install', 'setuptools==33.1.1')
+    run('pip', '--version')
+    run('sudo', 'make', 'install')
+  end
+when 'ruby'
+  run('gem', 'install', 'bundler', '-v', '1.5.1')
+else
+  abort("Missing or invalid SUITE environment variable")
+end
+
```
wvanbergen: The whitespace is off here.

PR #53, opened 2018-01-08
Fix Minitest 5.11 compatibility

casperisfine:
Minitest 5.11 was released a week ago, and changed the internals significantly breaking compatibility with most reporters out there (`minitest-reporters` included).

This PR test and support both 5.11 and previous versions.

### DIFF
--- a/ruby/lib/minitest/queue/runner.rb
+++ b/ruby/lib/minitest/queue/runner.rb
@@ -2,6 +2,7 @@ require 'optparse'
 require 'minitest/queue'
 require 'ci/queue'
 require 'digest/md5'
+require 'minitest/reporters/bisect_reporter'
 
 module Minitest
   module Queue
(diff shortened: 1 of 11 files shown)


ruby/test/support/reporter_test_helper.rb
```diff
@@ -3,6 +3,9 @@ module ReporterTestHelper
 
   def runnable(name, failure = nil)
     runnable = Minitest::Test.new(name)
+    if defined? Minitest::Result
+      runnable = Minitest::Result.from(runnable)
```
casperisfine: The main difference in 5.11 is here. Now instead of keeping all the run info inside the `Minitest::Test` instance that executed the test, it's copied inside a `Minitest::Result` instance. 

As such we need to use `Result#klass` to get the test class name.

ruby/lib/minitest/queue/runner.rb
```diff
@@ -130,7 +131,7 @@ def report_command
       def run_tests_in_fork(queue)
         child_pid = fork do
           Minitest.queue = queue
-          Minitest::Reporters.use!([Minitest::Reporters::SpecReporter.new])
+          Minitest::Reporters.use!([Minitest::Reporters::BisectReporter.new])
```
casperisfine: The minitest change basically broke `Minitest::Reporters::SpecReporter` so all the bisect integration test has a totally different output. SO while I was at it I implemented a simplified custom reporter for bisect.

PR #54, opened 2018-01-31
Refactor Minitest reporters to share code with future RSpec formatters

casperisfine:
This is the first of several refactors to make it easier to implement centralized reporting for RSpec.

The main goal here is that our custom minitest reporter stops talking to redis directly, but use a more high level interface that will be shared with RSpec. It has the added advantage of hiding wether we're using Redis, another backend, or a local in memory queue.


ruby/lib/ci/queue/build_record.rb
```diff
@@ -0,0 +1,44 @@
...
+        record_stats(stats)
+      end
+
+      def fetch_stats(stat_names)
+        stat_names.zip(stats.values_at(*stat_names).map(&:to_f))
+      end
+
+      private
+
+      attr_reader :stats
```
DazWorrall: Nitpick: could this be merged with the `attr_reader` at the top of the class?

ruby/lib/ci/queue/redis/build_record.rb
```diff
@@ -0,0 +1,63 @@
...
+          @config = config
+        end
+
+        def progress
+          @queue.progress
+        end
+
+        def queue_exhausted?
+          @queue.exhausted?
+        end
```
DazWorrall: There is some duplication in the methods above with `CI::Queue::BuildRecord`, could we inherit from it?

ruby/lib/ci/queue/build_record.rb
casperisfine: No because I want ti to be private.

ruby/lib/ci/queue/redis/build_record.rb
casperisfine: Yeah we could probably share a base class or a module. I'll keep that in mind of a future refactor.

PR #56, opened 2018-02-01
Centralized reporting for RSpec

casperisfine:
It can be improved in several ways, but this is kind of the MVP version.

Example use:

```bash
# Worker
rspec-queue --queue redis://localhost --build 1 --worker 1

# Summary
rspec-queue --queue redis://localhost --build 1 --report --timeout 180
```

### DIFF
--- a/ruby/lib/rspec/queue.rb
+++ b/ruby/lib/rspec/queue.rb
@@ -1,11 +1,40 @@
 require 'rspec/core'
 require 'ci/queue'
+require 'rspec/queue/build_status_recorder'
 
 module RSpec
   module Queue
     class << self
       def config
         @config ||= CI::Queue::Configuration.from_env(ENV)
+      end
+    end
+
+    module RunnerHelpers
+      private
+
+      def queue_url
+        configuration.queue_url || ENV['CI_QUEUE_URL']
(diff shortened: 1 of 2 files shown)


ruby/lib/rspec/queue.rb
```diff
@@ -142,8 +176,67 @@ def run(reporter)
...
+            abort! "No master was elected. Did all workers crash?"
+          end
+
+          unless supervisor.exhausted?
+            abort! "#{supervisor.size} tests weren't run."
+          end
+        end
+
+        # TODO: better reporting
+        errors = supervisor.build.error_reports.sort_by(&:first).map(&:last)
```
DazWorrall: Curious as to what the structure of  `error_reports` is that requires `.sort_by(&:first).map(&:last)`.

casperisfine: It's a `Hash[Example#id => String]`.

`Example#id` looks like: `./spec/foo_spec[1:1]`

The sort is so that errors coming from the same file are grouped together.

DazWorrall: Thanks for the explanation 👍

PR #57, opened 2018-02-09
Implement requeueing for RSpec

casperisfine:
The monkey patch is particularly tricky because RSpec internals don't expect a single `Example` to be ran twice. The `Example` instances are mutated to keep track of of the run result and then stored in various arrays for later printing the summary.

For example if you skip an example, its `execution_result.status` is set to `:pending` and the example instance is stored in `Reporter#pending_examples`. So if you reset the example instance to be able to run it again, it will stay in the `pending_examples` array and cause all sorts of trouble because it's state will be invalid for most reporters.

To work around this, when we requeue an example, we:
  - clone it 
  - reset the original example instance, and remove it from the reporter
  - Mark the clone as skipped, and add it to the reporter.

I tried to add a lot of comments to make the review easier for people not familiar with RSpec internal, but let me know if there is something you don't understand, as it might mean a comment is missing.

NB: There is a couple TODO left, mostly around making the output nicer, I'd rather handle them in a followup.

### DIFF
--- a/ruby/lib/rspec/queue.rb
+++ b/ruby/lib/rspec/queue.rb
@@ -1,3 +1,4 @@
+require 'delegate'
 require 'rspec/core'
 require 'ci/queue'
 require 'rspec/queue/build_status_recorder'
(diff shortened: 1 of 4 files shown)


ruby/lib/rspec/queue.rb
```diff
@@ -148,6 +149,54 @@ def queue_config
...
     RSpec::Core::Parser.prepend(ParserExtension)
 
+    module ExampleExtension
+      protected
+
+      def mark_as_requeued!(reporter)
+        @exception = nil
+        @metadata = @metadata.dup
+        @metadata[:execution_result] = execution_result.dup
+        execution_result.pending_message = "Failed but was requeued" # TODO: Improve message
```
wvanbergen: "The example failed, but another attempt will be done to rule out flakiness" ?

casperisfine: 👍

casperisfine: I'd like to see if I can include the failure details as well.

ruby/lib/rspec/queue.rb
```diff
@@ -148,6 +149,54 @@ def queue_config
 
     RSpec::Core::Parser.prepend(ParserExtension)
 
+    module ExampleExtension
+      protected
+
+      def mark_as_requeued!(reporter)
+        @exception = nil
+        @metadata = @metadata.dup
```
DazWorrall: This looks unnecessary to me, I'm probably missing something - could you explain?

casperisfine: Right.

So `Example` hold a `@metadata` `Hash`, and delegates someattributes to it: https://github.com/rspec/rspec-core/blob/7dd5edbf664a2213aabf21d06ebe537b8196089d/lib/rspec/core/example.rb#L48-L53

So `Example#execution_result` is actually `Example#@metadata[:execution_result]`. 

Since we want to mutate the result of the cloned example, without modifying to original one, we need to copy `@metadata` as well.

I'll add a comment to make this clearer.

casperisfine:
IIRC Sean is friend with an RSpec core team member, I'll ask around to see if they are interested looking at it. But as mentioned in the description, making our use case simpler would require quite a big change in the reporter / formatter API.

What worries me more is how easily it would break with further releases of RSpec.

PR #59, opened 2018-02-16
Embed a JUnit reporter for minitest

casperisfine:
This reporter is mostly taken from `minitest-reporters` but heavily simplified.

In a followup I'd like to extend it to include informations such as requeue etc.

### DIFF
--- a/ruby/.gitignore
+++ b/ruby/.gitignore
@@ -1,3 +1,4 @@
+test/fixtures/log/junit.xml
 /.byebug_history
 /.bundle/
 /.yardoc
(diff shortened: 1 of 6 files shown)


ruby/.gitignore
```diff
@@ -1,3 +1,4 @@
+test/fixtures/log/junit.xml
```
DazWorrall: DId you mean to ignore this? It looks like it's used in the tests below.

casperisfine: > Would this conflict with the reporter we currently use, or would the operate independently?

They'd work concurrently, however if they both try to write in the same path, one will win over the other. But there won't be name clashes or anything like that.

casperisfine: Yes, I want to ignore it because to written by the integration tests.

PR #62, opened 2018-02-20
Shopify's internal dev tool now requires a railgun.yml in addition to the dev.yml

mcgain:
I had to do this manually, which is odd, but dev now works properly again.


burke:
Ah, I probably only mass-migrated instances where the file lived at the top-level.

PR #63, opened 2018-02-26
Report on flaky tests separately

mcgain:
Read a file whose path has been provided as an ENV variable. That file
contains a list of flaky tests that should not be reported in the normal
fashion if they fail. Most reporters will mark them as skipped, but for
the JUnitXML reporter, we add a flaky_test attribute to the testCase xml
for the flaky tests.

I have not added this info to the readme, since it's really minimalistic at the moment andI'm not sure it's appropriate to add, or provided an rspec version.


ruby/lib/ci/queue/configuration.rb
```diff
@@ -1,21 +1,25 @@
+require 'ci/queue/null_flaky_supplier'
+require 'ci/queue/file_flaky_supplier'
```
casperisfine: I'd rather have those required in `ci/queue.rb`

ruby/lib/ci/queue/configuration.rb
```diff
@@ -1,21 +1,25 @@
...
-      attr_accessor :timeout, :build_id, :worker_id, :max_requeues, :requeue_tolerance, :namespace, :seed, :failing_test
+      attr_accessor :timeout, :build_id, :worker_id, :max_requeues, :requeue_tolerance, :namespace, :seed, :failing_test, :flaky_test_supplier
 
       class << self
         def from_env(env)
           new(
             build_id: env['CIRCLE_BUILD_URL'] || env['BUILDKITE_BUILD_ID'] || env['TRAVIS_BUILD_ID'],
             worker_id: env['CIRCLE_NODE_INDEX'] || env['BUILDKITE_PARALLEL_JOB'],
             seed: env['CIRCLE_SHA1'] || env['BUILDKITE_COMMIT'] || env['TRAVIS_COMMIT'],
+            flaky_test_supplier: (env['FLAKY_TEST_SUPPLIER_FILE'] && CI::Queue::FileFlakySupplier.new(env['FLAKY_TEST_SUPPLIER_FILE'])) || CI::Queue::NullFlakySupplier,
```
casperisfine: I think it would be better to namespace that variable. Something like `CI_QUEUE_FLAKY_TESTS`.

ruby/lib/ci/queue/file_flaky_supplier.rb
```diff
@@ -0,0 +1,18 @@
...
+
+      def initialize(filepath)
+        @test_list = JSON.parse(::File.read(filepath))
+        unless @test_list.is_a?(Array)
+          raise FileParseError, 'File must contain a JSON encoded array'
+        end
+      end
+
+      def include?(test)
+        @test_list.include?("#{test.class}##{test.name}")
```
casperisfine: This wouldn't work for RSpec. I think this method should either receive a `SingleExample` or a `SingleExample#id`

mcgain:
Feedback addressed. 

Passing a `SingleExample` into flaky turned out to be quite awkward, just passing the id is simpler. 

I also updated it to better match the existing env variable names and file formats.

ruby/lib/minitest/queue.rb
```diff
@@ -83,6 +94,11 @@ def queue_reporters=(reporters)
       @queue_reporters = reporters
     end
 
+    attr_reader :flaky_test_supplier
+    def flaky_test_supplier=(flaky_test_supplier)
```
casperisfine: Why not a simple `attr_accessor` ?

ruby/lib/ci/queue/file_flaky_supplier.rb
```diff
@@ -0,0 +1,19 @@
...
+    class FileFlakySupplier
+      FileParseError = Class.new(StandardError)
+
+      def initialize(filepath)
+        @test_list = ::File.readlines(filepath)
+        @test_list.map!(&:chomp)
+      end
+
+      def include?(runnable_id)
+        ::File.write('runnable_debuggoing.txt', "#{runnable_id} is looked for in #{@test_list}\n", mode: 'a')
```
casperisfine: Is this forgotten debug code?

ruby/lib/minitest/queue/runner.rb
```diff
@@ -134,6 +135,7 @@ def report_command
       def run_tests_in_fork(queue)
         child_pid = fork do
           Minitest.queue = queue
+          # Minitest.flaky_test_supplier = queue_config.flaky_test_supplier
```
casperisfine: Is this forgotten debug code as well?

ruby/lib/ci/queue/file_flaky_supplier.rb
```diff
@@ -0,0 +1,19 @@
+require 'json'
```
casperisfine: This is not longer used.

ruby/lib/minitest/queue.rb
casperisfine: Also I'd rather access `queue.flaky_test...` than increase the minitest patch surface.

We can even facade this with `queue.flaky?(test)`.

ruby/lib/minitest/queue.rb
```diff
@@ -50,6 +75,21 @@ def requeue!
...
+    def skipped?
+      super || flaked?
+    end
+
+    def flaked?
+      Flaked === failure
+    end
+
+    def mark_as_flaked!
+      self.failures.unshift(Flaked.new(self.failures.shift))
```
casperisfine: Just returning `true` to `skipped?` isn't enough, you also need `failures` to contain something with a particular `result_label` otherwise some parts of the Minitest API will still consider this result to be a failure (it doesn't expect this kind of extension TBH).

ruby/lib/minitest/queue.rb
```diff
@@ -103,6 +147,12 @@ def run_from_queue(reporter, *)
       queue.poll do |example|
         result = example.run
         failed = !(result.passed? || result.skipped?)
+
+        if failed && example.flaky?
+          result.mark_as_flaked!
```
casperisfine: We only mark as flaked in case of failure, we could do it all the time, I have no preference TBH. Thoughts?

ruby/lib/minitest/queue/junit_reporter.rb
```diff
@@ -59,7 +59,7 @@ def xml_message_for(test)
         xml = XmlMarkup.new(indent: 2, margin: 2)
         error = test.failure
 
-        if test.skipped?
+        if test.skipped? && !test.flaked?
```
casperisfine: Since the junit reporter is flaky test aware, it shouldn't mix actual skips and flakes.

ruby/test/integration/minitest_redis_test.rb
```diff
@@ -115,18 +116,18 @@ def test_junit_reporter
       end
       assert_empty err
       output = normalize(out.lines.last.strip)
-      assert_equal 'Ran 8 tests, 5 assertions, 1 failures, 1 errors, 1 skips, 3 requeues in X.XXs', output
+      assert_equal 'Ran 7 tests, 4 assertions, 1 failures, 1 errors, 1 skips, 2 requeues in X.XXs', output
```
casperisfine: Unless I'm mistaken, the changes here is because the flaky test isn't requeued anymore.

ruby/test/integration/minitest_redis_test.rb
```diff
@@ -136,25 +137,27 @@ def test_junit_reporter
...
          </testsuite>
-         <testsuite name="ATest" filepath="test/dummy_test.rb" skipped="3" failures="1" errors="0" tests="5" assertions="4" time="X.XX">
-           <testcase name="test_bar" lineno="8" classname="ATest" assertions="1" time="X.XX">
+         <testsuite name="ATest" filepath="test/dummy_test.rb" skipped="3" failures="1" errors="0" tests="4" assertions="3" time="X.XX">
+           <testcase name="test_bar" lineno="8" classname="ATest" assertions="1" time="X.XX" flaky_test="false">
            <skipped type="test_bar"/>
            </testcase>
-           <testcase name="test_flaky" lineno="12" classname="ATest" assertions="1" time="X.XX">
-           <skipped type="test_flaky"/>
+           <testcase name=\"test_flaky\" lineno=\"12\" classname=\"ATest\" assertions=\"1\" time=\"X.XX\" flaky_test=\"true\">
```
casperisfine: Subtle change here. Because of what I said above `flaky_test` is only `"true"` if it failed. Dunno if that's ok.

ruby/lib/minitest/queue.rb
mcgain: The external systems that determine whether something is flaky or not do have to keep track of this stuff, so there's no technical reason for us report it as flaky.

That said, passing flaky tests are still flaky so from a conceptual point of view, I guess we should still report it as flaky.

casperisfine: Indeed. I also just had some though: the junit integration in `junit` will have to be made aware as well, otherwise we'll display those failures as annotations.

mcgain: > the junit integration in junit will have to be made aware as well, otherwise we'll display those failures as annotations

I'm not sure what you mean here. Do you mean the messages in `lib/minitest/queue/junit_reporter.rb` or something else? If you do mean the junit reporter, we still need the failure messages for flaky tests.

mcgain: Oh, I see what you mean. I added some appropriate tests to demonstrate this. I suppose we'll need to create `Flaked` objects for the passes too.

mcgain:
In the JUnitXML, this will now mark the passing but still flaky tests as `flaky`. I first attempted to mirror the `Flaked` pattern, but that results in the wrong output, and it feels weird to produce failures when none actually exist.

My second attempt just uses an instance variable, and that results in the correct output. 

Could you please take another look.

casperisfine:
LGTM. Let's try this branch on a couple builds to sanity check.

mcgain:
A build that doesn't have any flaky tests in testability: https://buildkite.com/shopify/services-db/builds/435
a build that does: https://buildkite.com/shopify/billing/builds/400#_

here's one that uses the updated gem, but doesn't enable flaky_tests:  https://buildkite.com/shopify/services-db/builds/436

I couldn't test on shopify/shopify since it's using an older version of ci-queue that seems to need quite a bit of work to get it working. @DazWorrall could I have a review please.

DazWorrall:
> a build that does: https://buildkite.com/shopify/billing/builds/400#_

Interesting that the summary step reports no requeues 🤔 

2018-03-23 at 11 39

ruby/lib/minitest/queue.rb
```diff
@@ -35,6 +35,31 @@ def message
...
+  class Flaked < Skip
+    attr_reader :failure
+
+    def initialize(failure)
+      super()
+      @failure = failure
+    end
+
+    def result_label
+      "Skipped"
```
DazWorrall: Does this need to be `Skipped`?

mcgain: @byroot could you weigh in here, I think it's because this is for the other, non-junit reporters

mcgain:
> Interesting that the summary step reports no requeues 🤔

Requeues would be shown here if the test actually flaked, since it passed, it's not requeued. 

More concerning is that none of the tests in the junitxml are marked as flaky. Something's not right there.

ruby/lib/minitest/queue.rb
casperisfine: IIRC, no. It could be anything, but I can't be 100% sure what reporters do in the wild so I'd rather try to mimick skipped tests as much as possible

mcgain:
And I can't figure out why. I'll have a look tomorrow, but I might have to ask one of you to have a look if I can't figure it out quickly (I leave on vacation the day after tomorrow).

casperisfine:
I can take over if needed, but won't be able to focus 100% on this before a couple days.

mcgain:
Ugh, forgot I had a wedding on today, so I wasn't able to look at all. While this will do no harm if merged, updating shopify/shopify so the latest ci-queue works and testing on there seems the best route forwards. I really appreciate you taking a look whenever you get to it.

casperisfine:
I tested it, it seem to be working as expected. I'll merge and cut a new release.

PR #65, opened 2018-03-23
Add statsd reporter for Minitest CI Queue

jules2689:
Adds a statsd reporter extracted out of Shopify core. Includes a statsd implementation also extracted from shopify core.

This will allow stats to be submitted to analytics services so we can track flakiness and such over time.

It will always try to send stats, but the metrics won't submit without a `CI_QUEUE_STATSD_ADDR` env var being set


ruby/lib/minitest/queue/runner.rb
```diff
@@ -134,7 +135,10 @@ def report_command
       def run_tests_in_fork(queue)
         child_pid = fork do
           Minitest.queue = queue
-          Minitest::Reporters.use!([Minitest::Reporters::BisectReporter.new])
+          Minitest::Reporters.use!([
+            Minitest::Reporters::BisectReporter.new,
+            Minitest::Reporters::StatsdReporter.new,
```
casperisfine: I doubt you want the reporter here. This is for the bisect flow.

ruby/lib/minitest/queue/statsd.rb
```diff
@@ -0,0 +1,65 @@
+# Small, limited StatsD implementation.
+# We can't use the standard statsd-instrument for instrumenting our buildkite setup, because:
```
casperisfine: These comments are very weird considering we're in an open source gem.

We should rewrite them in a more general way.

ruby/test/minitest/reporters/order_reporter_test.rb
```diff
@@ -13,7 +13,7 @@ def test_before_test
       @reporter.before_test(runnable('a'))
       @reporter.before_test(runnable('b'))
       @reporter.report
-      assert_equal ['Minitest::Test#a', 'Minitest::Test#b'], File.readlines(log_path).map(&:chomp)
+      assert_equal ['Minitest::Result#a', 'Minitest::Result#b'], File.readlines(log_path).map(&:chomp)
```
casperisfine: This doesn't feel correct at all.

jules2689: There's some weird definitions going on in CI Queue:
https://github.com/Shopify/ci-queue/blob/master/ruby/lib/minitest/queue.rb#L119-L135

ruby/lib/minitest/queue/runner.rb
jules2689: Where do I register it instead?

ruby/test/minitest/reporters/order_reporter_test.rb
casperisfine: This is to provide compatibility between different versions of Minitest. That is why we test 2 different versions on CI.

ruby/lib/minitest/queue/runner.rb
casperisfine: Here I think: https://github.com/Shopify/ci-queue/blob/37c9de8ad2e96eb6f43b1538c2c182ba798a0de0/ruby/lib/minitest/queue/runner.rb#L46-L50

jules2689:
CI is failing because:
```
out, err = capture_subprocess_io do
        system(
          { 'BUILDKITE' => '1' },
          @exe, 'run',
          '--queue', @redis_url,
          '--seed', 'foobar',
          '--build', '1',
          '--worker', '1',
          '--timeout', '1',
          '--max-requeues', '1',
          '--requeue-tolerance', '1',
          '-Itest',
          'test/dummy_test.rb',
          chdir: 'test/fixtures/',
        )
      end
```

is trying to connect to statsd, but cannot connect.
How should we enable this? Should it be always enabled?

casperisfine:
> How should we enable this? Should it be always enabled?

I'd say env variable. The standard one is `STATSD_ADDR`, however since the tested app is likely to use it already, we should probably use `CI_QUEUE_STATSD_ADDR`, and simply not add the reporter if it's missing.

I'd also make sure connection failures & such have no effect. I wouldn't even print an error message TBH.

jules2689:
> I'd also make sure connection failures & such have no effect. 

Effect on what? I'm not sure what I'd check here.

casperisfine:
> Effect on what? I'm not sure what I'd check here.

Effect on the build. It would be silly to fail a build because we can't reach the statsd daemon. And spamming the build output wouldn't be great either.

jules2689:
re-ping @casperisfine

ruby/lib/minitest/queue/runner.rb
```diff
@@ -48,6 +49,7 @@ def run_command
           BuildStatusRecorder.new(build: queue.build),
           JUnitReporter.new,
           OrderReporter.new(path: 'log/test_order.log'),
+          Reporters::StatsdReporter.new,
```
casperisfine: I think it would better to check if statsd is configured here, and only add the reporter if it is.

Additionally, the `Configuration` object should be responsible for fetching the value from env.

jules2689:
can I get another quick look here @byroot ?

ruby/lib/minitest/queue/runner.rb
```diff
@@ -43,12 +44,16 @@ def retry_command
...
         set_load_path
         Minitest.queue = queue
-        Minitest.queue_reporters = [
+        reporters = [
           LocalRequeueReporter.new,
           BuildStatusRecorder.new(build: queue.build),
           JUnitReporter.new,
           OrderReporter.new(path: 'log/test_order.log'),
         ]
+        if @queue_config.statsd_endpoint
```
casperisfine: There is an `attr_reader` for `queue_config`

casperisfine:
I have to cut a new release with a fix today, so I took the liberty to directly fix my nitpicks on your PR, I'll merge on green and cut a release soon.

Issue #33, opened 2017-10-06
Allow to stop popping test if there already X failures


casperisfine:
In some extreme cases, a single low level issue can make a huge portion of the test suite fail. 

It can make sense in this context to stop running tests once X errors have been recorded, both to save CPU time and to keep a readable output.

cc @Shopify/pipeline

casperisfine:
NB: the tricky part is for the summary job to realize all workers gave up (when it happens).

wvanbergen:
We can use the new `wait` feature of BK, and only start the summarize job after all workers have finished even if some of them fail.

wvanbergen:
After #68, do we still need to do something for the summarize job?

casperisfine:
@DazWorrall suggested we could detect such issue by keeping track of the state of the queue. If nothing was popped for longer than the queue timeout then we could assume all workers are gone.

I think it would workin most case, but it's a bit finicky, because the queue could have kept the same size but test could have been requeued. It's quite a corner case though.

Alternatively we could have all workers write `Time.now` in a single key whenever they reach to Redis, this way the `report` command can see the last time a worker reported back.

In any case, I think It's fine to ship #68 without that safety in place. The report process waiting on dead workers is already an issue (e.g. all workers crash while loading the app), so I think we should treat it as a separate issue.

wvanbergen:
:+1: in the end the job timeout will fail the summary task as well

PR #68, opened 2018-04-06
Implement a circuit breaker to prevent unhealthy workers from exhausting requeue tickets

casperisfine:
Fix: https://github.com/Shopify/ci-queue/issues/33

### Context

In some cases, a worker process can end up in a bad state which makes pretty much every tests fail.

If the failure is slow to happen it's not much of a big deal as tests will be requeued and successfully ran by other workers. But sometimes the failure is very fast, in which case a single worker can pop and requeue almost the entire queue in a very short time.

### Solution

The `--max-consecutive-failure` option turns of a simplified [circuit breaker](https://en.wikipedia.org/wiki/Circuit_breaker_design_pattern), if the worker generates X failures in a row, the circuit will open and the worker will stop acquiring new jobs.

### Limitations

It's if all workers end up tripping the circuit breaker (e.g. all tests fail for a legit reason), then the report process (AKA summary step) won't beable to detect it and will hang until it reach it's timeout.

Ideally I'd like to detect such case but it's not easy, so I'd rather ship this as is and look at it later.


mcgain:
> Are you asking for some kind of error message to be displayed ?

Ah, should have been clearer, yes. An error message, some kind of explanation as to what has gone wrong and why we've stopped the build.

wvanbergen:
> Ah, should have been clearer, yes. An error message, some kind of explanation as to what has gone wrong and why we've stopped the build.

While I think a message may be nice for in order for us to debug things, I don't think that any end user will ever be confused about this. We already run a deterministic number of tests per worker, so  I don't think anybody will notice when a worker will shutdown early. The remaining tests will be covered by different workers.

Ideally, this will stop a worker while all the failures in that worker are still requeues, and we can still succeed the build.

mcgain:
I agree in the case where only one or two tests fail, but if someone pushes a branch with a syntax error (or, more likely, something else that legitimately causes a bunch of tests to fail), then all of the workers might end up in this situation and that's where messages are nice, otherwise I suspect we'll get questions like:
"I pushed this branch and I should have gotten way more errors than 10 per worker".

Ok, that's a pretty lame example, but I can imagine something similar coming up.

casperisfine:
I've added some shitty message to explain why the worker is exiting early. However there is no proper hook point for this in minitest so it's very dirty...

mrsimo:
Hi. Just wanted to bring up that @mcgain's comment happened to us today. We merged something that we didn't know would cause hundreds of errors, and all parallel workers succeeded with the "This worker is exiting early" message.

I guess I'm writing here in case other people end up in the same situation. Reading the output of the help I guess it makes a bit more sense:

> Defines after how many consecutive failures the worker will be considered unhealthy and terminate itself.

But I honestly thought it would fail early, not succeed early. I enabled it just because I thought it'd help clearly failing builds to fail quick and early without waiting for the full test suite to finish.

Are faulty workers a common thing for you? I'm surprised there's a whole feature for this problem.

byroot:
> But I honestly thought it would fail early, not succeed early.

Yeah, people are often surprised by this. The reasoning is that worker exit is mostly meaningless as they'll happily soft fail is anything's wrong. It's by design, workers have to be expandable so that infrastructure errors don't cripple builds.

Only the `report` exit code must be considered.

> Are faulty workers a common thing for you? I'm surprised there's a whole feature for this problem.

Very common yes. I understand that it might be surprising, but larger test suite face very different problems than smaller ones, and `ci-queue` is really optimized for huge ones. 

That being said, it seems the feature you'd like would be some kind of `--fail-fast`, as in shutdown the entire build after X confirmed failures happened. We don't have any use for that option, but I wouldn't be against merging such feature as I see how it could be helpful to some users.

mrsimo:
> Only the `report` exit code must be considered.

Ah... I didn't know that. We were only running `report` when there were failures. Would you say the correct way to run with `ci-queue` is to run `minitest-queue run` and then `minitest-queue report` on all nodes?

It would be nice if this was all documented somewhere. These behaviors are not always 100% straightforward, and https://github.com/Shopify/ci-queue/tree/master/ruby is quite minimalistic. I mean, `ci-queue` has been wonderful for us, and bless those requeues, it has saved us so much work with flakey tests. But this event almost gave me a heart attack when I saw I had deployed something with 90% of the test suite failing.

byroot:
> Would you say the correct way to run with `ci-queue` is to run `minitest-queue run` and then `minitest-queue report` on all nodes?

Nope. The correct way is to run `minitest-queue run` on all but one node, and `minitest-queue report` on the last one. Running it after all workers have completed is ok as well.

In the scenario you described (all workers soft failing), having `report` as a concurrent on sequential step would have prevented your build from succeeding, the reporter would have complained that not all tests were ran, and would have exited with `1`.

> It would be nice if this was all documented somewhere. 

Indeed, I actually tried but explaining these things is not my strong suit. I'll see if I can improve things, but no promises.

Issue #58, opened 2018-02-16
RSpec: failures are not acknowledged if a before hook fails


casperisfine:
```ruby
CI::Queue::Redis::ReservationError: "./spec/models/redacted_spec.rb[1:20]" is already reserved. You have to acknowledge it before you can reserve another one
--
  | /app/vendor/bundle/ruby/2.4.0/gems/ci-queue-0.11.1/lib/ci/queue/redis/worker.rb:117:in `reserve'
  | /app/vendor/bundle/ruby/2.4.0/gems/ci-queue-0.11.1/lib/ci/queue/redis/worker.rb:48:in `poll'
  | /app/vendor/bundle/ruby/2.4.0/gems/ci-queue-0.11.1/lib/rspec/queue.rb:347:in `block (2 levels) in run_specs'
  | /app/vendor/bundle/ruby/2.4.0/gems/rspec-core-3.6.0/lib/rspec/core/configuration.rb:1894:in `with_suite_hooks'
  | /app/vendor/bundle/ruby/2.4.0/gems/ci-queue-0.11.1/lib/rspec/queue.rb:346:in `block in run_specs'
  | /app/vendor/bundle/ruby/2.4.0/gems/rspec-core-3.6.0/lib/rspec/core/reporter.rb:79:in `report'
  | /app/vendor/bundle/ruby/2.4.0/gems/ci-queue-0.11.1/lib/rspec/queue.rb:343:in `run_specs'
  | /app/vendor/bundle/ruby/2.4.0/gems/rspec-core-3.6.0/lib/rspec/core/runner.rb:87:in `run'
  | /app/vendor/bundle/ruby/2.4.0/gems/rspec-core-3.6.0/lib/rspec/core/runner.rb:71:in `run'
  | /app/vendor/bundle/ruby/2.4.0/gems/rspec-core-3.6.0/lib/rspec/core/runner.rb:45:in `invoke'
  | /app/vendor/bundle/ruby/2.4.0/gems/ci-queue-0.11.1/exe/rspec-queue:4:in `<top (required)>'
  | /app/vendor/bundle/ruby/2.4.0/bin/rspec-queue:23:in `load'
  | /app/vendor/bundle/ruby/2.4.0/bin/rspec-queue:23:in `<top (required)>'


```

PR #73, opened 2018-05-29
Properly handle before suite hooks failing

casperisfine:
Fix: https://github.com/Shopify/ci-queue/issues/58

Long story short, when a `before :suite` hooks raises:

  - `@world.non_example_failure` is set to true
  - `@world.wants_to_quit` is set to true

If this happens, we don't even want to start polling, otherwise other parts of RSpec will bail out early and the global state will be inconsistent.


DazWorrall:
Hmm, try adding `--ignore-installed six` to https://github.com/Shopify/ci-queue/blob/82c1581895ff6b3ef16ca05054478797085ce502/python/Makefile#L30

PR #80, opened 2018-06-29
rspec-queue: prevent early exit from exiting with a failure status

casperisfine:
Some jobs are exiting with status `1` even though all the examples they ran succeeded, and the report job succeed.

I suspect it's the `return if RSpec.world.wants_to_quit` which is causing this by returning `nil` from `SingleExample#run`.

### DIFF
--- a/ruby/dev.yml
+++ b/ruby/dev.yml
@@ -3,7 +3,7 @@
 name: ci-queue
 
 up:
-- ruby: 2.1.6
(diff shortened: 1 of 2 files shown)


ianks:
I believe this is causing our CI to pass when in actuality, no tests actually run. This can happen in config errors, gem loading errors etc. This is a bit issue for us. Can this be merged?

PR #81, opened 2018-06-29
Add support for default Heroku CI ENV vars and REDIS_URL

schneems:
### Heroku CI

Details on Heroku env vars are found at:

- https://devcenter.heroku.com/articles/heroku-ci#environment-variables-env-key
- https://devcenter.heroku.com/articles/heroku-ci-parallel-test-runs#parallelizing-your-tests

### REDIS_URL

Similar to `DATABASE_URL` the `REDIS_URL` is a semi-standard at this point, it’s also what you get by default if you add a Redis provider to Heroku. 

When no env var is set and no option given we should default to the default local installation port of redis.

BTW this project is amazing and you are amazing ❤️ I used this software to get my CI runs down from 9 minutes to 5 minutes: https://github.com/heroku/heroku-buildpack-ruby/pull/784#issuecomment-400841723 

The test invocation is at https://github.com/heroku/heroku-buildpack-ruby/pull/784/files#diff-6a1b6963e828b69fcc667e3c45df78bbR23.

### DIFF
--- a/ruby/lib/minitest/queue/runner.rb
+++ b/ruby/lib/minitest/queue/runner.rb
@@ -198,7 +198,7 @@ module Minitest
           help = split_heredoc(<<-EOS)
             Unique identifier for the workload. All workers working on the same suite of tests must have the same build identifier.
             If the build is tried again, or another revision is built, this value must be different.
-            It's automatically inferred on Buildkite, CircleCI and Travis.
+            It's automatically inferred on Buildkite, CircleCI, Heroku CI, and Travis.
           EOS
           opts.separator ""
           opts.on('--build BUILD_ID', *help) do |build_id|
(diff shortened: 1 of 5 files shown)


ruby/lib/minitest/queue/runner.rb
```diff
@@ -314,7 +315,7 @@ def ordering_seed
       end
 
       def queue_url
-        @queue_url || ENV['CI_QUEUE_URL']
+        @queue_url || ENV['CI_QUEUE_URL'] || ENV['REDIS_URL'] || "redis://localhost:6379"
```
casperisfine: `$REDIS_URL` is indeed a semi-standard, however it is much more likely `$REDIS_URL` will point a a Redis used by the tested app and not `ci-queue`.

It's actually a common setup error we got, people pointing ci-queue to the Redis instance dedicated to their test worker instead of the global one.

I'm also not sure defaulting to `localhost:6379` is a good idea, as it's very unlikely to be a working value.

casperisfine:
Also I'm curious, how did you provision a shared redis instance in Heroku CI ?

schneems:
I moved the Heroku CI env vars PR over here https://github.com/Shopify/ci-queue/pull/82 and the `REDIS_URL` pr over here: https://github.com/Shopify/ci-queue/pull/83

PR #83, opened 2018-07-02
Default to REDIS_URL if present fallback localhost

schneems:
Similar to `DATABASE_URL` the `REDIS_URL` is a semi-standard at this point, it’s also what you get by default if you add a Redis provider to Heroku. 

When no env var is set and no option given we should default to the default local installation port of redis.

Moved only the REDIS_URL stuff from #81 to this PR.

### DIFF
--- a/ruby/lib/minitest/queue/runner.rb
+++ b/ruby/lib/minitest/queue/runner.rb
@@ -188,7 +188,8 @@ module Minitest
 
           help = split_heredoc(<<-EOS)
             URL of the queue, e.g. redis://example.com.
-            Defaults to $CI_QUEUE_URL if set.
+            Will use $CI_QUEUE_URL or $REDIS_URL if set.
+            Defaults to redis://localhost:6379.
           EOS
           opts.separator ""
           opts.on('--queue URL', *help) do |url|
@@ -314,7 +315,7 @@ module Minitest
       end
 
       def queue_url
-        @queue_url || ENV['CI_QUEUE_URL']
+        @queue_url || ENV['CI_QUEUE_URL'] || ENV['REDIS_URL'] || "redis://localhost:6379"
       end
 
       def invalid_usage!(message)
--- a/ruby/lib/minitest/queue/runner.rb
+++ b/ruby/lib/minitest/queue/runner.rb
@@ -188,8 +188,7 @@ module Minitest
 
           help = split_heredoc(<<-EOS)
             URL of the queue, e.g. redis://example.com.
-            Will use $CI_QUEUE_URL or $REDIS_URL if set.
-            Defaults to redis://localhost:6379.
+            Defaults to $CI_QUEUE_URL or $REDIS_URL if set.
           EOS
           opts.separator ""
           opts.on('--queue URL', *help) do |url|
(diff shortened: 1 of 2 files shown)


casperisfine:
> When no env var is set and no option given we should default to the default local installation port of redis.

I'll just repeat my opinion from #81. I think it's very likely that if a `REDIS_URL` is present, it's actually a Redis that is for the actual application use and isn't shared.

I'd rather have the command fail, and people explicitly pass `--queue "${REDIS_URL}"`, than having it behave weirdly (every workers running the full suite).

For context, I added `CI_QUEUE_URL` so that we could offer 1st class support of `ci-queue` on our private CI (we have a shared Redis instance and automatically set that env var for all build).

schneems:
Replying to @casperisfine from #81

> Also I'm curious, how did you provision a shared redis instance in Heroku CI ?

You can set it in your `app.json` right here under "addons": https://github.com/heroku/heroku-buildpack-ruby/pull/784/files#diff-6a1b6963e828b69fcc667e3c45df78bbR7

> $REDIS_URL is indeed a semi-standard, however it is much more likely $REDIS_URL will point a a Redis used by the tested app and not ci-queue.

I've not got as much operational experience here, what's the harm in running both in the same redis? Do people do things like `flushdb` regularly in their tests? It looks like these test items auto get a namespace based off of the build ID. Maybe we could also default to adding an extra namespace element to help people differentiate? Any extra context here is helpful.

Based on "It's actually a common setup error we got, people pointing ci-queue to the Redis instance dedicated to their test worker instead of the global one." it sounds like people are provisioning redis inside of the test node itself?

One possible work around here could be adding support for metadata, such as total number of nodes in the CI. We could then maybe have each node "check in" via redis and if the number does not match what we think it should be raise some kind of error. For example Heroku gives you `CI_NODE_TOTAL`. I know that would be a little more difficult for something like travis where not all nodes will start at the beginning. However maybe it just raises an error if 0 nodes "register" after a certain time?

It sounds like this is already a problem today and would benefit from some usability love. 

> I'm also not sure defaulting to localhost:6379 is a good idea, as it's very unlikely to be a working value.

That's fine. I removed it as the default and instead I added it to the error message so if someone intended to run with it, they can copy and paste from the error message without having to look it up on google.

casperisfine:
> I've not got as much operational experience here, what's the harm in running both in the same redis? Do people do things like flushdb regularly in their tests?

`flushdb` is indeed a concern. But more generally when you parallelize your tests, each test process has it's own set of services (RDBMS / Redis / Memcache / etc). So all workers don't have the same `REDIS_URL`.

>  it sounds like people are provisioning redis inside of the test node itself?

Yes, that is exactly the scenario for which we implemented `ci-queue`. In our case each test workers is a Docker container, with a set of supporting services launched with Docker compose. Whereas the Redis instance behind `$CI_QUEUE_URL` is a global one for the entire test cluster.

> One possible work around here could be adding support for metadata...  if the number does not match what we think it should be raise some kind of error.

The problem with this approach is that `ci-queue` is built to that if all workers but one crash on boot, the last one will run all the tests and have a chance to succeed. This is because we run our tests with a huge number of workers (400 at this point) so even if a very small fraction of it fail on boot, it cause lots of builds to fail for non-relevant reasons.

> It sounds like this is already a problem today and would benefit from some usability love.

Indeed. The `report` process could do what you suggest, e.g. if it doesn't see any worker come up after X time, it could error out and show a nice error message.

> they can copy and paste from the error message without having to look it up on google.

Yup totally. If we had a nice detection mechanism for that issue, then I'd be fine with using `REDIS_URL` as default.

PR #86, opened 2018-07-11
Do not exit 1 when RSpec decided to exit early

casperisfine:
Under some conditions, RSpec might decide to exit without running all the tests (e.g. got SIGINT).

When it happens it sets `RSpec.world.wants_to_quit = true`, which it checked in several places and cause early returns / breaks.

In this case `Example#run` will return `nil` instead of the expected `true/false`. Before this PR we were not expecting it, so `nil` was acting as `false`.

The default RSpec behavior in this case is to exit with status `1`, but in our case we know other workers will pick up the slack, so we can safely exit 0 if all the test we ran were successes.

The added integration test showcases this scenario.


schneems:
I haven't tried with this change, but previously I accidentally left a require in my spec helper file after removing the gem. When I did that, all the jobs failed because the require didn't exist but the tests still exited 0. Should I open a new issue for this?

```
----> Running test command `HATCHET_BUILDPACK_BRANCH=$HEROKU_TEST_RUN_BRANCH bundle exec rspec-queue --max-requeues=3 --timeout 180 --build $HEROKU_TEST_RUN_ID --worker $CI_NODE_INDEX --queue $REDIS_URL`...
An error occurred while loading ./spec/hatchet/bugs_spec.rb.
Failure/Error: require 'knapsack'
LoadError:
  cannot load such file -- knapsack
# ./spec/spec_helper.rb:9:in `require'
# ./spec/spec_helper.rb:9:in `<top (required)>'
# ./spec/hatchet/bugs_spec.rb:1:in `require_relative'
# ./spec/hatchet/bugs_spec.rb:1:in `<top (required)>'
# ./vendor/bundle/ruby/2.4.0/gems/ci-queue-0.13.4/lib/rspec/queue.rb:363:in `setup'
An error occurred while loading ./spec/hatchet/ci_spec.rb.
Failure/Error: require 'knapsack'
LoadError:
  cannot load such file -- knapsack
# ./spec/spec_helper.rb:9:in `require'
# ./spec/spec_helper.rb:9:in `<top (required)>'
# ./spec/hatchet/ci_spec.rb:1:in `require'
# ./spec/hatchet/ci_spec.rb:1:in `<top (required)>'
# ./vendor/bundle/ruby/2.4.0/gems/ci-queue-0.13.4/lib/rspec/queue.rb:363:in `setup'
```

casperisfine:
@schneems it often confuse people, but it's actually expected behavior. Let me try to explain:

When you run your test suite with very high parallelism, it's not unlikely that a few jobs won't be able to boot because of various reasons, e.g. some datastore being unresponsive, the node they run on being unhealthy etc.

As such a test worker not booting is considered to be a soft failure.

In case they don't boot because of an actual real programming error, then the `--report` step will notice no test was run and will fail.

PR #87, opened 2018-07-27
Only run failed tests on retry

casperisfine:
### Previous retry behavior

The worker would simply run the exact same list of tests it ran before failing / crashing

### New retry behavior

The worker only run the failed test it ran previously. 

If no test failed, then it simply goes back to poll the general queue.

If the general queue is exhausted as well, it early exit.

### Rationale

The previous behavior was implemented as is for consistency reasons, if a test was impacted by a leak we wanted the test order to be the same so that it's replicable.

But really retries happen for two reasons:

  - The developer manually retry the job because it failed due to a flaky or leaky test. They actually want the leak not to happen again, so we might as well only retry failures so that it's both faster and less likely to be impacted by the leak.

  - The job was retried automatically because we detected an infrastructure failure, in this case re-running successful tests is wasteful, we might as go back to poll the test queue.

### DIFF
--- a/ruby/test/fixtures/test/passing_test.rb
+++ b/ruby/test/fixtures/test/passing_test.rb
@@ -1,0 +1,9 @@
+require 'test_helper'
+
+class PassingTest < Minitest::Test
+  100.times do |i|
+    define_method("test_passing_#{i}") do
+      assert true
+    end
+  end
+end
--- a/ruby/lib/ci/queue/redis/build_record.rb
+++ b/ruby/lib/ci/queue/redis/build_record.rb
@@ -16,6 +16,10 @@ module CI
           @queue.exhausted?
         end
 
+        def failed_tests
+          redis.hkeys(key('error-reports'))
+        end
+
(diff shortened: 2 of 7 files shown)


ruby/test/fixtures/test/passing_test.rb
```diff
@@ -0,0 +1,9 @@
+require 'test_helper'
+
+class FailingTest < Minitest::Test
```
mcgain: class name should be `PassingTest`?

casperisfine: Yup, some lazy copy paste here... /o\

ruby/test/fixtures/test/passing_test.rb
```diff
@@ -0,0 +1,9 @@
+require 'test_helper'
+
+class PassingTest < Minitest::Test
+  100.times do |i|
+    define_method("test_passing_#{i}") do
+      assert true
```
wvanbergen: Nit: you can do `pass` instead of `assert true`

PR #88, opened 2018-09-26
Reset minitest stats on retry

casperisfine:
In some circumstances, after retrying failed workers a build could end up with the report step complaining about errors, but no error reports not failed workers.

What would happen is that a worker would have failed, but it's failing test would have been retried by another worker. So all test ultimately pass.

However when retrying the first worker, it sees the queue is all worked off and simply bail out with a success.

The problem is that the report job was looking at `failures` and `errors` stats, which would have been left pristine by the retried worker.

The solution here is to look at the `error-reports` hash instead, which has the benefit of being updated in an idempotent way, so is a much better source of truth.

Additionally, workers reset their minitest stats on retry, this will cause less consistent stats, but make it much less likely to have a report steps succeeded while displaying a number of `errors/failures` which will inevitably alert users.

### DIFF
--- a/ruby/lib/minitest/queue/build_status_reporter.rb
+++ b/ruby/lib/minitest/queue/build_status_reporter.rb
@@ -25,7 +25,7 @@ module Minitest
       end
 
       def success?
-        errors == 0 && failures == 0
+        build.error_reports.empty?
       end
 
       def record(*)
--- a/ruby/test/fixtures/test/flaky_test.rb
+++ b/ruby/test/fixtures/test/flaky_test.rb
@@ -1,0 +1,11 @@
+require 'test_helper'
+
+class FlakyTest < Minitest::Test
+  def test_dummy
+    assert true
+  end
+
+  def test_flaky
+    assert_equal '1', ENV['FLAKY_TEST_PASS']
+  end
+end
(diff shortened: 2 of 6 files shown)


ruby/lib/minitest/queue/build_status_reporter.rb
```diff
@@ -25,7 +25,7 @@ def report
       end
 
       def success?
-        errors == 0 && failures == 0
+        build.error_reports.empty?
```
casperisfine: The most important change is here. All the rest is mostly for display / stats purpose.

ruby/test/fixtures/test/flaky_test.rb
```diff
@@ -0,0 +1,11 @@
...
+
+class FlakyTest < Minitest::Test
+  def test_dummy
+    assert true
+  end
+
+  def test_flaky
+    assert_equal '1', ENV['FLAKY_TEST_PASS']
+  end
+end
```
dylanahsmith: Is this dead code?  Since the added test uses `test/failing_test.rb`.  Looks like you might have tried to use this to `Simulate another worker successfuly retrying all errors` but forgot to remove this code when you couldn't get that working.

casperisfine: Yes, totally dead code. I'll 🔥

PR #89, opened 2018-10-01
Flush `OrderReporter` before running test

DazWorrall:
Small change so we can always trust that, while running, the value in the logs is up to date.


DazWorrall:
> we could change the proctitle to make it easy to see what test is being processed.

Nice idea, those can be easily truncated through layers of docker/ssh/etc though.

PR #90, opened 2018-10-09
Rename Shopify Build Env Vars

JackTLi:
Adding in new shopify build env vars

This PR was generated by a script.

### DIFF
--- a/ruby/lib/ci/queue/configuration.rb
+++ b/ruby/lib/ci/queue/configuration.rb
@@ -10,7 +10,7 @@ module CI
           new(
             build_id: env['CIRCLE_BUILD_URL'] || env['BUILDKITE_BUILD_ID'] || env['TRAVIS_BUILD_ID'] || env['HEROKU_TEST_RUN_ID'],
             worker_id: env['CIRCLE_NODE_INDEX'] || env['BUILDKITE_PARALLEL_JOB'] || env['CI_NODE_INDEX'],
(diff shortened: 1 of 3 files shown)


ruby/lib/ci/queue/configuration.rb
```diff
@@ -10,7 +10,7 @@ def from_env(env)
           new(
             build_id: env['CIRCLE_BUILD_URL'] || env['BUILDKITE_BUILD_ID'] || env['TRAVIS_BUILD_ID'] || env['HEROKU_TEST_RUN_ID'],
             worker_id: env['CIRCLE_NODE_INDEX'] || env['BUILDKITE_PARALLEL_JOB'] || env['CI_NODE_INDEX'],
-            seed: env['CIRCLE_SHA1'] || env['BUILDKITE_COMMIT'] || env['TRAVIS_COMMIT'] || env['HEROKU_TEST_RUN_COMMIT_VERSION'],
+            seed: env['CIRCLE_SHA1'] || env['SHOPIFY_BUILD_COMMIT'] || env['TRAVIS_COMMIT'] || env['HEROKU_TEST_RUN_COMMIT_VERSION'],
```
casperisfine: In this case we should support both. `ci-queue` is a public gems, some people might use it with raw Buildkite.

JackTLi: Ah yes 👍

PR #92, opened 2018-12-13
nextitem + no shuffling

simon-weber:
Here's two changes we've been running in prod for a while. I was lazy and opened them together since I wasn't sure if the python plugin was active anyway.

Nearly everything is to support pytest's `nextitem`: passing it can improve performance significantly by avoiding unnecessary fixture teardown/setup (xdist does the same thing, and has some notes on it [here](https://github.com/pytest-dev/pytest-xdist/blob/master/OVERVIEW.md)). Essentially, the workers now try to pull two tests instead of one, and finish once the queue is empty (to avoid deadlocks when eg there is one test left and multiple workers with only one buffered locally).

I also tweaked ItemIndex to not implicitly order tests by their hash value in python 2.

### DIFF
--- a/python/ciqueue/distributed.py
+++ b/python/ciqueue/distributed.py
@@ -70,7 +70,7 @@ class Worker(Base):
 
     def __iter__(self):
         def poll():
-            while not self.shutdown_required and len(self):  # pylint: disable=len-as-condition
+            while not self.shutdown_required and self.redis.llen(self.key('queue')):
                 test = self._reserve()
                 if test:
                     yield test.decode()
--- a/python/ciqueue/pytest.py
+++ b/python/ciqueue/pytest.py
@@ -174,8 +174,28 @@ def pytest_runtestloop(session):
         config.pluginmanager.register(RedisReporter(config, queue))
     session.items = ItemList(tests_index, queue)
 
-    for item in session.items:
-        item.config.hook.pytest_runtest_protocol(item=item, nextitem=None)
+    items = iter(session.items)
+
+    # Pull two tests at a time to provide nextitem to pytest, which
+    # allows for optimal teardown of fixtures.
+    try:
+        item = next(items)
+    except StopIteration:
+        # There were no tests queued.
+        return True
+
+    while True:
+        try:
+            nextitem = next(items)
+        except StopIteration:
+            # There are no more tests -- finish our final one.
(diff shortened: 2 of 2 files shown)


python/ciqueue/distributed.py
```diff
@@ -70,7 +70,7 @@ def __init__(self, tests, worker_id, redis, build_id,  # pylint: disable=too-man
 
     def __iter__(self):
         def poll():
-            while not self.shutdown_required and len(self):  # pylint: disable=len-as-condition
+            while not self.shutdown_required and self.redis.llen(self.key('queue')):
```
simon-weber: there's a subtle change here: workers no longer wait for in-progress tests to finish once the queue is empty. The chance of this causing issues seems unlikely, but I figured I'd point it out.

I don't think there's a super easy fix without putting more information in redis, but I might be wrong.

casperisfine: > workers no longer wait for in-progress tests to finish once the queue is empty.

That was on purpose. If they don't wait, then there won't be workers ready to process requeues.

simon-weber: yup, makes sense. That was a hack to prevent a deadlock where multiple workers pull 1 item, then wait forever on the others to finish.

What do you think of this:
* revert the llen change
* modify the lua polling script to pop 2 items when possible, or 1 if it's the last item
* when receiving 1 item, workers process it immediately (ie nextitem=None)

I think that'll fix the deadlock, since there should be no situation where workers are waiting on an item that's not proceeding. It wastes some fixture time in the case where one of the final tests needs to be retried, but that seems unlikely enough to not be a big deal (and it shouldn't affect correctness).

casperisfine: Hum, I'm not sure I understand how you ended up with a dead lock. But I'm fine with the proposed lua change as long as it's an option, because it doesn't offer any benefit to the ruby implementation.

simon-weber: Cool. To clarify, here's an example timeline:
* queue is empty, worker A and B each holding one item
* worker A polls to buffer a second item. redis returns nothing, and since worker B's item may be requeued, worker A continues polling
* worker B does the same thing, since worker A's buffered item may be requeued

I'll look into the lua solution when I have a chance.

PR #94, opened 2019-01-17
Deprecate the retry subcommand and make the behavior automatic

casperisfine:
The `retry` subcommand (or `--retry` flag for rspec) makes it hard to integrate on CI, and we realized many of our pipeline didn't integrate that part.

This caused and imperfect behavior that surprises people.

Since not all CI systems expose jobs retry as an env var, I chose to inspect Redis to detect wether we are retrying or not.

### DIFF
--- a/ruby/lib/minitest/queue/runner.rb
+++ b/ruby/lib/minitest/queue/runner.rb
@@ -37,15 +37,22 @@ module Minitest
       end
 
       def retry_command
-        reset_counters
-        retry_queue = queue.retry_queue
-        unless retry_queue.exhausted?
-          self.queue = retry_queue
-        end
-        run_command
+        STDERR.puts "Warning: the retry subcommand is deprecated."
+        run_command # aliased for backward compatibility purpose
       end
 
       def run_command
+        if queue.retrying?
+          reset_counters
+          retry_queue = queue.retry_queue
+          if retry_queue.exhausted?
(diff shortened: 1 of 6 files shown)


ruby/lib/minitest/queue/runner.rb
```diff
@@ -37,15 +37,22 @@ def run!
...
+        STDERR.puts "Warning: the retry subcommand is deprecated."
+        run_command # aliased for backward compatibility purpose
       end
 
       def run_command
+        if queue.retrying?
+          reset_counters
+          retry_queue = queue.retry_queue
+          if retry_queue.exhausted?
+            puts "The retry queue does not contain any failure, we'll process the main queue instead."
```
casperisfine: I figured some output on the selected behavior would help future debugging. Feel free to suggest alternative phrasing as english is not my strong suit.

casperisfine:
Note: The python CI is failing, I'm not sure why yet, but whatever it is, it's not due to this PR, so please disregard.

ruby/lib/minitest/queue/runner.rb
DazWorrall: Some possible alternatives:

> The retry queue does not contain any failures, we'll process the main queue instead.

> Found 0 tests to retry, processing the main queue.

> This worker has no previous test failures, so nothing to retry. Processing the main queue.
>

PR #95, opened 2019-02-28
Add support for Semaphore 2

gregnavis:
Shopify, thank you for publishing the gem!

I'd like to contribute support for Semaphore 2 with the following defaults:

- Build ID - taken from [`SEMAPHORE_PIPELINE_ID`](https://docs.semaphoreci.com/article/12-environment-variables#semaphore_pipeline_id) (but see my note below)
- Worker ID - taken from [`SEMAPHORE_JOB_ID`](https://docs.semaphoreci.com/article/12-environment-variables#semaphore_job_id)
- Seed - based on [`SEMAPHORE_GIT_SHA`](https://docs.semaphoreci.com/article/12-environment-variables#semaphore_git_sha)

Regarding build ID, I wasn't sure whether to chose `SEMAPHORE_PIPELINE_ID` or [`SEMAPHORE_WORKFLOW_ID`](https://docs.semaphoreci.com/article/12-environment-variables#semaphore_workflow_id). According to [the documentation](https://docs.semaphoreci.com/article/62-concepts), a workflow is comprised of several pipelines which are comprised of several blocks.

My thinking is that by scoping `ci-queue` to a pipeline we'll have stronger isolation in cases when two pipelines running in parallel use `ci-queue`. However, I'm not 100% sure on that and would like your feedback here.

### DIFF
--- a/ruby/README.md
+++ b/ruby/README.md
@@ -24,6 +24,7 @@ Or install it yourself as:
   - CircleCI
   - Travis
   - Heroku CI
+  - Semaphore 2
 
 If you are using another CI system, please refer to the command usage message.
 
--- a/ruby/lib/ci/queue/configuration.rb
+++ b/ruby/lib/ci/queue/configuration.rb
@@ -8,9 +8,9 @@ module CI
       class << self
         def from_env(env)
           new(
-            build_id: env['CIRCLE_BUILD_URL'] || env['BUILDKITE_BUILD_ID'] || env['TRAVIS_BUILD_ID'] || env['HEROKU_TEST_RUN_ID'],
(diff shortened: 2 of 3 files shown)


casperisfine:
> Regarding build ID, I wasn't sure whether to chose `SEMAPHORE_PIPELINE_ID` or [`SEMAPHORE_WORKFLOW_ID`](https://docs.semaphoreci.com/article/12-environment-variables#semaphore_workflow_id).

Reading the doc I would says pipeline makes more sense, but I could see use case for doing it at the workflow level. If people wnat to do that they can always overwrite the value by passing another env var.

so 👍

PR #96, opened 2019-03-05
Remove reset_counters method for rspec

DougEdey:
I bumped `ci-queue` to 0.15.0 for spy, but because it uses rspec I found that the `reset_counters` wasn't in the rspec queue.

EDIT: And is not needed.

I kinda reverse engineered this a bit, but tried to keep it inline with the minitest queue. 

~I'm not sure how to test it either :( I'm not a huge rspec person~ Added test

https://buildkite.com/shopify/spy/builds/10719#39e0272c-08d9-43e9-a75a-afdee9935c23/312-318

Also fixed Python tests

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -47,7 +47,7 @@ setuplib.setup(
         'test': [
             'tox==2.9.1',
             'shopify_python==0.4.1',
-            'pycodestyle == 2.3.0',
+            'pycodestyle == 2.4.0',
         ]
     },
     package_data={'': get_lua_scripts()},
--- a/ruby/lib/rspec/queue.rb
+++ b/ruby/lib/rspec/queue.rb
@@ -377,7 +377,6 @@ module RSpec
             puts "Found 0 tests to retry, processing the main queue."
           else
             puts "Retrying #{retry_queue.size} failed tests."
-            reset_counters
             queue = retry_queue
           end
         end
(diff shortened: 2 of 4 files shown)


ruby/lib/rspec/queue/build_status_recorder.rb
```diff
@@ -3,6 +3,8 @@ module Queue
     class BuildStatusRecorder
       ::RSpec::Core::Formatters.register self, :example_passed, :example_failed
 
+      COUNTERS = ['error-reports']
```
casperisfine: Not sure this constant is necessary.

casperisfine:
> I'm not sure how to test it either :( I'm not a huge rspec person

For minitest it's tested as an integration test: https://github.com/Shopify/ci-queue/blob/79ea69dc7ea1866838fa7c4f8ac6ddb7b4c78e8a/ruby/test/integration/minitest_redis_test.rb#L145

ruby/lib/rspec/queue/build_status_recorder.rb
DougEdey: Turns out there's no need to reset counters at all with rspec, so I added a test and removed the `reset_counters` method entirely

DougEdey: Because the try deletes the original key on success: https://github.com/Shopify/ci-queue/blob/master/ruby/lib/ci/queue/redis/build_record.rb#L37

Issue #98, opened 2019-04-15
Allow to reproduce order-related failures by printing a full executable command


elia:
When an order-related failure happens is very difficult to reproduce it locally for large suites (which are likely among users of this library).

Having a way to reproduce the exact sequence locally would be a great help and would considerably reduce `--bisect` times.

_If there's consensus I can try to propose an implementation for this 🙋‍♂️_

casperisfine:
I don't know which adapter and language version you use, but this is supported in most of them.

With minitest for instance, the default reporters will write the test sequence in `log/test_order.log`, and you can use that to replay it with `minitest-queue --queue path/to/test_order.log`.

casperisfine:
That being said, the whole thing might be a bit finicky, and any PR polishing or documenting that flow is welcome.

elia:
@casperisfine we're using rspec, I'll have a look at test_order.log and how to use it with rspec, then send a PR with all the documentation I'll be able to gather 👍

casperisfine:
Then yeah, that feature is missing from the RSpec integration.

schneems:
I just came to ask for such a feature. I'm also using rspec. The Heroku Ruby Buildpack has an intermittent test that has been failing for months now, usually, I restart the test and it works but, I would like to dig in deeper.

Looks like that feature is implemented in minitest here https://github.com/Shopify/ci-queue/blob/e9bcff63a7f62afd750417ca2c5b1f5289abee6b/ruby/lib/minitest/queue/runner.rb#L62. Do you know if there's a fundamental difficulty to implementing something like it with rspec or if it's just not been done yet?

byroot:
> Do you know if there's a fundamental difficulty to implementing something like it with rspec or if it's just not been done yet?

I haven't investigated it yet, so no idea if there is such difficulty. But from vague memory of RSpec internals I think it should be relatively easy. IIRC RSpec already have a syntax for invoking tests in a specific order, and it even have built-in bisect. Also `File` queues expose the same interface as `Redis` queues, so `rspec-queue --queue path/to/test_order.log` should already work today.

So what need to be done is to log the test identifiers in a `test_order.log` file.

PR #99, opened 2019-06-06
Implement test_order logging for RSpec implementation

casperisfine:
Fix: https://github.com/Shopify/ci-queue/issues/98

Log test order in a file with the RSpec implementation to be able to replay the sequence with `rspec-queue --queue path/to/test_order.log`.

@schneems

### DIFF
--- a/ruby/.gitignore
+++ b/ruby/.gitignore
@@ -11,4 +11,4 @@ test/fixtures/log/test_order.log
 /spec/reports/
 /tmp/
 /lib/ci/queue/redis/*.lua
-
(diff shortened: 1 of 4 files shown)


schneems:
I can confirm it works for me. Here's my branch https://github.com/heroku/heroku-buildpack-ruby/pull/890.

Since Heroku CI doesn't have "artifacts" i'm having to output to the STDOUT instead:

```
<ci-queue-command-here> || { cat log/test_order.log;  $(exit 1); }
```

PR #101, opened 2019-06-27
Grind command

doodzik:
This is a draft for a grind command, which enables us to check if a test is flaky by running it over and over again. In the current version of ci-queue we are able to run a test multiple times, but we aren't able to summarize it. We require a summary step as it gives us a way to communicate the grinding results in a more understandable way. 

When we try to use the report functionality the previous test runs get overridden by newer ones, which causes the reporting to be wrong. We isolated the problem down to tests not being uniquely identifiable. We solve it by adding a sequential id at the end of every test name.

This solution works, but the changes made to `id` are pretty invasive and will lead to confusion. Do you think there is a better way to solving this problem?

here is an example run of the working grind command and reporter:

https://buildkite.com/shopify/shopify-build-onboarding-testing/builds/727#a9a7823d-09a3-4a84-9e48-0a93afb60ddf

\cc @mcgain

### DIFF
--- a/ruby/lib/ci/queue/grind.rb
+++ b/ruby/lib/ci/queue/grind.rb
@@ -1,0 +1,22 @@
+require 'ci/queue/static'
+
+module CI
+  module Queue
+    class Grind < Static
+      class << self
+        def from_uri(uri, config)
+          new(uri.path, config)
+        end
+      end
+
+      def initialize(path, config)
+        io = path == '-' ? STDIN : ::File.open(path)
+
+        tests_to_run = io.each_line.map(&:strip).reject(&:empty?)
+        test_grinds = (tests_to_run * config.grind_count).sort
+
+        super(test_grinds, config)
+      end
+    end
+  end
+end
--- a/ruby/lib/minitest/queue/grind_recorder.rb
+++ b/ruby/lib/minitest/queue/grind_recorder.rb
@@ -1,0 +1,67 @@
+module Minitest
+  module Queue
+    class GrindRecorder < Minitest::Reporters::BaseReporter
+
+      attr_accessor :test_count
+
+      def self.counters
+        @counters ||= {
+          'failures' => 0,
+          'errors' => 0,
+          'skips' => 0,
+          'test_count' => 0
+        }
+      end
+
+      class << self
+        attr_accessor :failure_formatter
+      end
+      self.failure_formatter = FailureFormatter
+
+      def initialize(build:, **options)
+        super(options)
+        @build = build
+      end
+
+      def record(test)
+        increment_counter(test)
+        record_test(test)
+      end
+
+      private
+
+      def record_test(test)
+        stats = self.class.counters
+        if (test.failure || test.error?) && !test.skipped?
+          build.record_error(dump(test), stats: stats)
+        else
+          build.record_success(stats: stats)
+        end
+      end
+
+      def increment_counter(test)
+        if test.skipped?
+          self.class.counters['skips'] += 1
+        elsif test.error?
+          self.class.counters['errors'] += 1
+        elsif test.failure
+          self.class.counters['failures'] += 1
+        end
+        self.class.counters['test_count'] += 1
+
+        key = "count##{test.klass}##{test.name}"
+
+        unless self.class.counters.key?(key)
+          self.class.counters[key] = 0
+        end
+        self.class.counters[key] += 1
+      end
+
+      def dump(test)
+        ErrorReport.new(self.class.failure_formatter.new(test).to_h).dump
+      end
+
+      attr_reader :build
+    end
+  end
+end
--- a/ruby/lib/minitest/queue/runner.rb
+++ b/ruby/lib/minitest/queue/runner.rb
@@ -70,11 +70,43 @@ module Minitest
         trap('INT') { Minitest.queue.shutdown! }
 
         if queue.rescue_connection_errors { queue.exhausted? }
-          puts green("All tests were ran already")
+          puts green('All tests were ran already')
         else
           load_tests
           populate_queue
         end
+        # Let minitest's at_exit hook trigger
+      end
+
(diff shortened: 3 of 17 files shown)


casperisfine:
> This solution works, but the changes made to `id` are pretty invasive and will lead to confusion. Do you think there is a better way to solving this problem?

Yeah, I don't like it either. Did you explore an alternative storage structure?

The main BuildRecorder uses a Hash to make it easy to override a previous failure, but in case of grind you simply want to append the failures reports to a list. So a Redis list might make more sense.

ruby/lib/minitest/queue/grind_recorder.rb
```diff
@@ -45,9 +45,9 @@ def record(test)
 
         stats = self.class.counters
         if (test.failure || test.error?) && !test.skipped?
-          build.record_error("#{test.klass}##{test.name}##{SecureRandom.uuid}", dump(test), stats: stats)
```
doodzik: @mcgain I think this still has to be included. I'm going to play around with this.

mcgain:
@casperisfine could you please take another look? 

Darren and Frederik ended up pairing on this and came up with the approach of using a Static Queue, which works much better than the id mutating implementation. 

Still todo in this PR:
* I'd like to remove the inheritance in the `GrindSupervisor` and `GrindRecord` by injecting the build into their current parents, then we can just ditch those classes altogether. 
* Squash into a better narrative
* add a test around how this grind interacts with other commands, to prove they don't interfere with each other when running concurrently. 

But a review on the direction and implementation would be good.

ruby/lib/minitest/queue/runner.rb
```diff
@@ -70,14 +70,50 @@ def run_command
...
+      def grind_command
+        invalid_usage!('No list to grind provided') if grind_list.nil?
+        invalid_usage!('No grind count provided') if grind_count.nil?
+
+        set_load_path
+
+        queue_config.build_id = queue_config.build_id + '-grind'
+
+        reporter_queue = CI::Queue::Redis::Grind.new(queue_url, queue_config)
+        @queue = CI::Queue::Static.new([], queue_config)
```
casperisfine: Why this? It's not used and overwritten later.

ruby/lib/minitest/queue/runner.rb
```diff
@@ -70,14 +70,50 @@ def run_command
...
+        if queue_config.statsd_endpoint
+          reporters << Minitest::Reporters::StatsdReporter.new(statsd_endpoint: queue_config.statsd_endpoint)
+        end
+        Minitest.queue_reporters = reporters
+
+        trap('TERM') { Minitest.queue.shutdown! }
+        trap('INT') { Minitest.queue.shutdown! }
+
+        load_tests
+        tests_to_run = File.readlines(grind_list).map(&:strip!).to_set
```
casperisfine: Why can't `Queue::File` be used here? It seems to be doing pretty much the same thing.

Worst case I'd rather see some subclass of it if really necessary.

ruby/lib/ci/queue/redis/worker.rb
```diff
@@ -21,7 +21,7 @@ def initialize(redis, config)
 
         def populate(tests, random: Random.new)
           @index = tests.map { |t| [t.id, t] }.to_h
-          tests = Queue.shuffle(tests, random)
+          tests = Queue.shuffle(tests, random) if random.class == Random
```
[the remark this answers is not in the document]
doodzik: I removed this all-together as we aren't using the redis queue anymore

mcgain:
I squashed the commits, I think a single one is sufficient. 

@doodzik had a go at removing the inheritance, but felt the result was more complicated than what we have now, so I'll leave it as it. 

I didn't get around to adding that test, but manual testing shows that there is no interference.

ruby/lib/ci/queue/grind.rb
```diff
@@ -0,0 +1,26 @@
...
+        def from_uri(uri, config)
+          new(uri.path, config)
+        end
+      end
+
+      def initialize(path, config)
+        io = path == '-' ? STDIN : ::File.open(path)
+
+        tests_to_run = io.each_line.map(&:strip).reject(&:empty?).to_set
+        tests = Minitest.loaded_tests.select do |test|
```
casperisfine: You shouldn't access `Minitest` stuff in here.

There is 3 main section in ci queue:

  - `lib/minitest` the minitest integration
  - `lib/rspec` the rspec integration
  - `lib/ci` the abstract code that backs both.

Anyway, a Queue implementation should be initialized with string identifiers, the mapping to actual test framework objects is done later via `populate`.

doodzik: 👍 thanks for the clarification. I changed the code appropriately.

ruby/lib/ci/queue/grind.rb
```diff
@@ -0,0 +1,22 @@
...
+        def from_uri(uri, config)
+          new(uri.path, config)
+        end
+      end
+
+      def initialize(path, config)
+        io = path == '-' ? STDIN : ::File.open(path)
+
+        tests_to_run = io.each_line.map(&:strip).reject(&:empty?).to_set
+        test_grinds = Array.new(config.grind_count) { tests_to_run.to_a }.flatten.sort
```
casperisfine: So the implementation is very close to `File`.

If you inherited you could simply do:

```ruby
super
@queue *= config.grind_count
```

doodzik: The one thing that makes me a little bit hesitant is that the Static class sets `@total = tests.size` in its initializer, which the File class inherits from. I think that would complicate things, but I might be wrong.

casperisfine: RIght. I would at least simplify the building of the repetition. That `Array.new` and the `.to_set` is much more complex that it needs to.

PR #102, opened 2019-07-09
add ability to exit grind after a specified timeout is reached

doodzik:
This is a follow up PR to #101. It introduces a flag that allows to specify the max duration that a worker should be able to work off its queue.

We want to be able to exit a grind after a certain time as grinding a slow test can slow down a pipeline significantly. Slower tests tend to be more flaky as they depend on other services to run. Therefor running them not as often shouldn't affect the effectiveness of the grind.

The main difference to a regular timeout is that it isn't intended to make the worker fail.


ruby/lib/ci/queue/circuit_breaker.rb
```diff
@@ -13,6 +13,37 @@ def report_success!
...
+
+        def report_success!
+        end
+
+        def open?
+          @closes_at < current_timestamp
+        end
+
+        def message
+          'This worker is exiting early because it reached its timeout'
```
wvanbergen: Should we include the timeout duration in the message?

wvanbergen:
Can you please leave a comment w.r.t. the use case, as we discussed in the pod?

mcgain:
Willem's comment was resolved offline

doodzik:
> I am not sure what scenarios this timeout is trying to prevent, but you may have to use a thread to work around that issue.

How this timeout works is that after the specified time passed we will not pull new tests from the queue. It isn't a strict timeout as we don't need that. Our main issue is that we need to prevent the a slow test can clog up our test onboarding pipeline. Let's say we have a test that takes 1 minute, but we specify that we grind it 600 times. This will result in a very slow pipeline that takes 10h. We can gracefully stop grind the grind after X minutes to prevent that issue. 

> So the issue with this approach is that if the code is stuck somewhere, we will never ask if the circuit breaker is open, so we will not exit this after a timeout.

We have an `--timeout` flag for that use case

ruby/lib/ci/queue/configuration.rb
```diff
@@ -39,14 +40,19 @@ def initialize(
         @flaky_tests = flaky_tests
         @statsd_endpoint = statsd_endpoint
         @grind_count = grind_count
+        @max_duration = max_duration
```
casperisfine: You got a bug here. Since you don't go though the setter, `CircuitBreaker::Timeout` won't be appended to the list.

doodzik: opened a pr with the fix. Thanks :)

PR #103, opened 2019-07-11
fix max duration setter

doodzik:
We aren't using the the max_duration flag setter in the configuration initializer. This prevents it to be added to the circuit breakers.

ref #102


doodzik:
> Would be good to see a regression test for this.

I think in this case it isn't necessary as it didn't cause any misbehaviour when running the software.
Thus the benefit of having a specific test for this case isn't really there.

PR #104, opened 2019-08-19
Use exception class for error, failure and skipped type

### LINKED ISSUE #77: Failure type in JUnit XML makes no sense
I think this is causing the issue where skipped tests are reported in the buildkite annotation, instead of any actual failures:

```xml
<testsuite name="ShopSearchIndexingJobTest" filepath="components/platform/test/unit/jobs/shop_search_indexing_job_test.rb" skipped="1" failures="0" errors="0" tests="1" assertions="0" time="0.06969359999993685">
  <testcase name="test_update_shop_documents_when_shop's_active_field_changes" lineno="30" classname="ShopSearchIndexingJobTest" assertions="0" time="0.06969359999993685" flaky_test="true">
    <failure type="test_update_shop_documents_when_shop's_active_field_changes" message="This test is very slow: [url]">
Skipped: test_update_shop_documents_when_shop's_active_field_changes(ShopSearchIndexingJobTest) [/app/components/platform/test/unit/jobs/shop_search_indexing_job_test.rb:31]: This test is very slow: [url]
    </failure>
  </testcase>
</testsuite>
```

This was a skipped test (as in: `skip` at the top of the test definition, not testability magic). 2 things:

1. The failure type makes no sense (it's the test name)
1. Possibly because of the above, you can't tell from the XML that is was skipped. It looks like a regular failure, so our parsing logic can't ignore these

cc @Shopify/pipeline

dylanahsmith:
cc @DazWorrall 
Fixes #77

## Problem

Using the test name for the error or failure type didn't make sense, wasn't aligned with the [Apache Ant's JUnit XML schema description](https://github.com/windyroad/JUnit-Schema/blob/master/JUnit.xsd) of the error type and wasn't useful by redundantly including the test name again in this field.

The exception class wasn't previously included in the junit.xml and I would like to have to it to help determine the cause of a test failure when it is due to a timeout or infrastructure error.  It can also be useful to determine whether a test was actually skipped or was just retried.

## Solution

Use the exception class name for the type of the skipped, failure and error elements.

### DIFF
--- a/ruby/lib/minitest/queue/junit_reporter.rb
+++ b/ruby/lib/minitest/queue/junit_reporter.rb
@@ -57,16 +57,16 @@ module Minitest
 
       def xml_message_for(test)
         xml = XmlMarkup.new(indent: 2, margin: 2)
-        error = test.failure
+        failure = test.failure
 
         if test.skipped? && !test.flaked?
-          xml.skipped(type: test.name)
+          xml.skipped(type: failure.error.class.name)
         elsif test.error?
-          xml.error(type: test.name, message: xml.trunc!(error.message)) do
+          xml.error(type: failure.error.class.name, message: xml.trunc!(failure.message)) do
             xml.text!(message_for(test))
           end
-        elsif test.failure
-          xml.failure(type: test.name, message: xml.trunc!(error.message)) do
+        elsif failure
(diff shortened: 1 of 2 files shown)


casperisfine:
If you are confident, then ship it.

PR #107, opened 2019-10-29
add output test failure results in a seperate command

doodzik:
This PR writes the test failures into a file so we can inspect them easily and calculate metrics on top of them.

This PR also fixes an issue with the python ci

ref: https://github.com/Shopify/test-infra/issues/10


ruby/test/integration/minitest_redis_test.rb
```diff
@@ -329,6 +329,68 @@ def test_junit_reporter
       END
     end
 
+    def test_redis_reporter_failure_file
+      failure_file = File.expand_path('../../fixtures/tmp/failure_file.json', __FILE__)
```
esnunes: Based on the build errors the `tmp` folder doesn't seem to exist. Have you considered using [Tempfile](https://ruby-doc.org/stdlib-2.4.0/libdoc/tempfile/rdoc/Tempfile.html) or [mktmpdir](https://ruby-doc.org/stdlib-2.5.3/libdoc/tmpdir/rdoc/Dir.html#method-c-mktmpdir)?

doodzik:
I probably should have looked at your comments earlier. I've ended up bumping the py.test gem. It also turned out that git doesn't track empty directories, which is why ci failed.

ruby/test/integration/minitest_redis_test.rb
```diff
@@ -329,6 +330,59 @@ def test_junit_reporter
       END
     end
 
+    def test_redis_reporter_failure_file
+      Dir.mktmpdir do |dir|
+        failure_file = File.join(dir, 'failure_file.json')
+
+        out, err = capture_subprocess_io do
```
sander-m: `out, err` are unused variables. Should we add an assertion on them (e.g. `assert_empty err`)?

sander-m:
Do we need to bump the [gem version](https://github.com/Shopify/ci-queue/blob/master/ruby/lib/ci/queue/version.rb#L3) as part of this change?

PR #108, opened 2019-11-07
Add py37 and py38 to `tox`

imnotjames:
To validate that we can run Python 3.7 and 3.8 on ciqueue let's add them to the tox runner!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,10 @@ addons:
       - deadsnakes
     packages:
       - python3.6
+      - python3.7
+      - python3.7-distutils
+      - python3.8
+      - python3.8-distutils
 services:
   - redis-server
 language: ruby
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -5,6 +5,10 @@ addons:
       - deadsnakes
     packages:
       - python3.6
+      - python3.7
+      - python3.7-distutils
+      - python3.8
+      - python3.8-distutils
```
doodzik: why do we need to add `python3.8-distutils` for python 3.8 and 3.7, but not for 3.6

imnotjames: There is no `python3.6-distutils` package - for some reason it seems to have `distutils` included.  See related travis CI job: https://travis-ci.org/Shopify/ci-queue/jobs/609006985

However, without `distutils` installed in the standard library, `tox` chokes up https://travis-ci.org/Shopify/ci-queue/jobs/609003518

PR #110, opened 2019-11-13
Extract opt parse for quick review

jessicaxiejw:
move option parse out of runner.rb because
- reducing the scope of runner.rb
- the new ParseEnv and ParseArgv make it easier to see where a config can come from


doodzik:
In general this looks like a good cleanup.
I’m not sure if I would go with the ParseArgv=ParseEnv here. That sounds more like an action rather than a class. I think it is more that we have a Configuration class that takes an argv and/or env.
In any case the parsing should live under the `Configuration` namespace.

ruby/lib/minitest/queue/parse_env.rb
```diff
@@ -0,0 +1,34 @@
+module CI
```
doodzik: I think something like this would be cleaner here:
```ruby
class Configuration
    EnvConfig = Struct.new(
        :build_id,
        :worker_id,
        :seed,
        :statsd_endpoint,
        :queue_url,
        :flaky_tests
      ) do

      def self.parse(env)
        config = new()
        config.build_id = env['CIRCLE_BUILD_URL'] || env['BUILDKITE_BUILD_ID'] || env['TRAVIS_BUILD_ID'] || env['HEROKU_TEST_RUN_ID'] || env['SEMAPHORE_PIPELINE_ID']
        config.worker_id = env['CIRCLE_NODE_INDEX'] || env['BUILDKITE_PARALLEL_JOB'] || env['CI_NODE_INDEX'] || env['SEMAPHORE_JOB_ID']
        config.seed = env['CIRCLE_SHA1'] || env['BUILDKITE_COMMIT'] || env['TRAVIS_COMMIT'] || env['HEROKU_TEST_RUN_COMMIT_VERSION'] || env['SEMAPHORE_GIT_SHA']
        config.statsd_endpoint = env['CI_QUEUE_STATSD_ADDR']
        config.queue_url = ENV['CI_QUEUE_URL']
        config.flaky_tests = load_flaky_tests(env['CI_QUEUE_FLAKY_TESTS'])
        config
      end

      private

      def self.load_flaky_tests(path)
        return [] unless path
        ::File.readlines(path).map(&:chomp).to_set
      rescue SystemCallError
        []
      end
    end
end
```

ruby/lib/minitest/queue/parse_env.rb
```diff
@@ -0,0 +1,34 @@
...
+        :flaky_tests
+      )
+
+      def self.parse(env)
+        config = Config.new()
+        config.build_id = env['CIRCLE_BUILD_URL'] || env['BUILDKITE_BUILD_ID'] || env['TRAVIS_BUILD_ID'] || env['HEROKU_TEST_RUN_ID'] || env['SEMAPHORE_PIPELINE_ID']
+        config.worker_id = env['CIRCLE_NODE_INDEX'] || env['BUILDKITE_PARALLEL_JOB'] || env['CI_NODE_INDEX'] || env['SEMAPHORE_JOB_ID']
+        config.seed = env['CIRCLE_SHA1'] || env['BUILDKITE_COMMIT'] || env['TRAVIS_COMMIT'] || env['HEROKU_TEST_RUN_COMMIT_VERSION'] || env['SEMAPHORE_GIT_SHA']
+        config.statsd_endpoint = env['CI_QUEUE_STATSD_ADDR']
+        config.queue_url = ENV['CI_QUEUE_URL']
```
doodzik: this ENV is capitalized. It probably shouldn't be.

ruby/lib/minitest/queue/parse_env.rb
doodzik: Same goes for the other ArgvConfig

doodzik: While you do this refactor it is probably good to rename `CI_QUEUE_FLAKY_TESTS ` to `CI_QUEUE_INTERMITTENT_TESTS` and deprecate the former one. (Not getting rid of it so we don't break existing usage.)

ruby/lib/minitest/queue/parse_argv.rb
```diff
@@ -0,0 +1,200 @@
...
+        @argv = argv
+        @config = Config.new
+      end
+
+      def config
+        parser.parse!(@argv)
+        @config
+      end
+
+      def parser
```
doodzik: I think each flag should get it's own method. That should aid readability a little bit.

ruby/lib/ci/queue/configuration.rb
```diff
@@ -1,81 +1,78 @@
...
+        @config.failure_file = argv_config.failure_file
+        @config.circuit_breakers = [CircuitBreaker::Disabled]
+        @config.max_requeues = argv_config.max_requeues || 0
+        @config.requeue_tolerance = argv_config.requeue_tolerance || 0
+        @config.queue_url = argv_config.queue_url || env_config.queue_url
+        @config.flaky_tests = env_config.flaky_tests || []
+        @config.timeout = argv_config.timeout || 30
+        build_id = fetch_build_id(argv_config.namespace, env_config.build_id, argv_config.build_id)
+        @config.build_id = build_id
+        @config.seed = argv_config.seed || env_config.seed || build_id
```
doodzik: We are duplicating logic here from Argv and Env config. It might be good to just access them directly from those objects.

ruby/lib/minitest/queue/runner.rb
```diff
@@ -19,15 +18,16 @@ def self.invoke(argv)
       end
 
       def initialize(argv)
-        @queue_config = CI::Queue::Configuration.from_env(ENV)
-        @command, @argv = parse(argv)
+        @queue_config = CI::Queue::Configuration.new(ENV, argv).config
+        @command = argv[0]
+        @files = argv[1..-1]
       end
```
sander-m: The old code looks more readable. Don't you think so?

jessicaxiejw: I can put it behind a new method. However, the old method is mutating the argv object, which is dangerous.

PR #112, opened 2019-11-18
foo

jessicaxiejw:


ruby/lib/minitest/queue/runner.rb
```diff
@@ -402,6 +412,15 @@ def parser
             queue_config.max_consecutive_failures = Integer(max)
           end
 
+          help = <<~EOS
+            Set the time limit on the median of the execution time from grinds on a given test.
+            The unit is seconds and decimal is allowed.
+            Defaults to disabled.
+          EOS
+          opts.on('--median-execution-time-limit LIMIT_IN_MILLISECONDS', *help) do |limit|
```
jessicaxiejw: change to max-test-duration, add max-test-duration-percentile

ruby/lib/minitest/queue/runner.rb
```diff
@@ -402,6 +413,32 @@ def parser
...
+
+          help = <<~EOS
+            The percentile for max-test-duration. For example, when max-test-duration is set to 10 and
+            max-test-duration-percentile is set to 0.5, the test's median execution time during a grind must be
+            lower than 10 milliseconds.
+            The percentile must be within the range 0 < percentile <= 1.
+            Defaults to 0.5 (50th percentile).
+          EOS
+          opts.on('--max-test-duration-percentile LIMIT_IN_MILLISECONDS', *help) do |percentile|
+            queue_config.max_test_duration_percentile = Float(percentile)
```
doodzik: why not `to_f`?

ruby/lib/minitest/queue/runner.rb
```diff
@@ -197,9 +199,18 @@ def report_grind_command
...
+
+        test_time_record = CI::Queue::Redis::TestTimeRecord.new(queue_url, queue_config)
+        test_time_reporter = Minitest::Queue::TestTimeReporter.new(
+          build: test_time_record,
+          limit: queue_config.max_test_duration,
+          percentile: queue_config.max_test_duration_percentile,
+        )
+        test_time_reporter.report
+
+        exit! grind_reporter.success? && test_time_reporter.success? ? 0 : 1
```
doodzik: As we add another option to fail here I was thinking that it would be good for the grind reporter to give us some feedback on what failed that is parseable by a machine. This would give us the ability to collect metrics on how often things fail. Exporting the result as a json would be fine. I was thinking about an exit code but that would just hide more complex situations where multiple things fail.

ruby/lib/minitest/queue/runner.rb
jessicaxiejw: I forgot about .to_f :P I will update it.

ruby/lib/minitest/queue/runner.rb
jessicaxiejw: Yeah, I was thinking of something similar. Right now, a test takes too long, the message "Do you think you have intermittently failing tests?" pop up in shopify core.

PR #113, opened 2019-11-20
Add TestTimeRecord

jessicaxiejw:
Adding the class to store the run time of individual test during a grind in Redis.

Related https://github.com/Shopify/test-infra/issues/36

You can see the entire change at https://github.com/Shopify/ci-queue/pull/112/files

### DIFF
--- a/ruby/lib/ci/queue/redis/test_time_record.rb
+++ b/ruby/lib/ci/queue/redis/test_time_record.rb
@@ -30,7 +30,7 @@ module CI
         def record_test_name(test_name)
           redis.pipelined do
             redis.lpush(
-              key_to_list_all_test_names,
+              all_test_names_key,
               test_name.force_encoding(Encoding::BINARY),
             )
           end
@@ -39,7 +39,7 @@ module CI
 
         def fetch_all_test_names
           values = redis.pipelined do
-            redis.lrange(key_to_list_all_test_names, 0, -1)
+            redis.lrange(all_test_names_key, 0, -1)
           end
           values.flatten.map(&:to_s)
         end
@@ -52,7 +52,7 @@ module CI
           values.flatten.map(&:to_f)
         end
 
-        def key_to_list_all_test_names
+        def all_test_names_key
           "build:#{config.build_id}:list_of_test_names".force_encoding(Encoding::BINARY)
         end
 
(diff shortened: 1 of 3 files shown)


ruby/lib/ci/queue/redis/test_time_record.rb
```diff
@@ -0,0 +1,65 @@
...
+          values.flatten.map(&:to_f)
+        end
+
+        def key_to_list_all_test_names
+          "build:#{config.build_id}:list_of_test_names".force_encoding(Encoding::BINARY)
+        end
+
+        def test_time_key(test_name)
+          "build:#{config.build_id}:#{test_name}".force_encoding(Encoding::BINARY)
+        end
```
sander-m: The method names `test_time_key ` and `key_to_list_all_test_names` are somewhat inconsistent.

ruby/test/ci/queue/redis/test_time_record_test.rb
```diff
@@ -0,0 +1,27 @@
+require 'test_helper'
+
+class CI::Queue::Redis::TestTimeRecordTest < Minitest::Test
+  def setup
+    redis_url = "redis://#{ENV.fetch('REDIS_HOST', 'localhost')}/7"
+    redis = Redis.new(url: redis_url)
+    redis.flushdb
+
+    config ||= CI::Queue::Configuration.new(
```
sander-m: Why do we use `||=` here?

ruby/test/ci/queue/redis/test_time_record_test.rb
```diff
@@ -0,0 +1,27 @@
+require 'test_helper'
+
+class CI::Queue::Redis::TestTimeRecordTest < Minitest::Test
+  def setup
+    redis_url = "redis://#{ENV.fetch('REDIS_HOST', 'localhost')}/7"
+    redis = Redis.new(url: redis_url)
+    redis.flushdb
```
sander-m: Nitpick: this same pattern is used in multiple places, it would be nice to move this into a test helper module that any test can include if it needs a local Redis instance (more DRY). Feel free to ignore this for this PR.

jessicaxiejw: good idea, I'll add it in a separate PR.

doodzik: Can you create an issue for that and add it to the kanban board, so that this doesn't get lost?

jessicaxiejw: good idea, issue created: https://github.com/Shopify/ci-queue/issues/114

PR #116, opened 2019-11-29
add --max-test-duration feature

jessicaxiejw:
Add `--max-test-duration` and `--max-test-duration-percentile` to grind.

Related https://github.com/Shopify/test-infra/issues/36

You can see the entire change at https://github.com/Shopify/ci-queue/pull/112/files

### DIFF
--- a/ruby/lib/minitest/queue/runner.rb
+++ b/ruby/lib/minitest/queue/runner.rb
@@ -89,10 +89,12 @@ module Minitest
         queue_config.grind_count = grind_count
 
         reporter_queue = CI::Queue::Redis::Grind.new(queue_url, queue_config)
+        test_time_record = CI::Queue::Redis::TestTimeRecord.new(queue_url, queue_config)
 
         Minitest.queue = queue
         reporters = [
-          GrindRecorder.new(build: reporter_queue.build)
+          GrindRecorder.new(build: reporter_queue.build),
+          TestTimeRecorder.new(build: test_time_record)
         ]
         if queue_config.statsd_endpoint
           reporters << Minitest::Reporters::StatsdReporter.new(statsd_endpoint: queue_config.statsd_endpoint)
@@ -197,9 +199,18 @@ module Minitest
           abort! error.message
         end
 
-        reporter = GrindReporter.new(build: supervisor.build)
-        reporter.report
-        exit! reporter.success? ? 0 : 1
+        grind_reporter = GrindReporter.new(build: supervisor.build)
+        grind_reporter.report
+
+        test_time_record = CI::Queue::Redis::TestTimeRecord.new(queue_url, queue_config)
+        test_time_reporter = Minitest::Queue::TestTimeReporter.new(
+          build: test_time_record,
+          limit: queue_config.max_test_duration,
+          percentile: queue_config.max_test_duration_percentile,
+        )
+        test_time_reporter.report
+
+        exit! grind_reporter.success? && test_time_reporter.success? ? 0 : 1
       end
 
       private
@@ -276,7 +287,7 @@ module Minitest
             Defaults to $CI_QUEUE_URL if set.
           EOS
           opts.separator ""
-          opts.on('--queue URL', *help) do |url|
+          opts.on('--queue URL', help) do |url|
             self.queue_url = url
           end
 
@@ -284,7 +295,7 @@ module Minitest
             Path to the file that includes the list of tests to grind.
           EOS
           opts.separator ""
-          opts.on('--grind-list PATH', *help) do |url|
+          opts.on('--grind-list PATH', help) do |url|
             self.grind_list = url
           end
 
@@ -292,7 +303,7 @@ module Minitest
             Count defines how often each test in the grind list is going to be run.
           EOS
           opts.separator ""
-          opts.on('--grind-count COUNT', *help) do |count|
+          opts.on('--grind-count COUNT', help) do |count|
             self.grind_count = count.to_i
           end
 
@@ -302,7 +313,7 @@ module Minitest
             It's automatically inferred on Buildkite, CircleCI, Heroku CI, and Travis.
           EOS
           opts.separator ""
-          opts.on('--build BUILD_ID', *help) do |build_id|
+          opts.on('--build BUILD_ID', help) do |build_id|
             queue_config.build_id = build_id
           end
 
@@ -311,7 +322,7 @@ module Minitest
               Example: --namespace integration
           EOS
           opts.separator ""
-          opts.on('--namespace NAMESPACE', *help) do |namespace|
+          opts.on('--namespace NAMESPACE', help) do |namespace|
             queue_config.namespace = namespace
           end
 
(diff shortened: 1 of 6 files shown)


ruby/test/integration/grind_redis_test.rb
```diff
@@ -188,5 +187,124 @@ def test_can_grind_multiple_things
...
+        '--worker', '1',
+        '--timeout', '1',
+        '--grind-count', '10',
+        '--grind-list', 'grind_list_success.txt',
+        '-Itest',
+        'test/dummy_test.rb',
+        chdir: 'test/fixtures/',
+      )
+
+      out, err = capture_subprocess_io do
```
sander-m: `out` is unused. Should we assert it's empty?  (if not `_out` makes sense).

ruby/lib/minitest/queue/test_time_reporter.rb
```diff
@@ -0,0 +1,69 @@
+require 'minitest/reporters'
+
+module Minitest
+  module Queue
+    class TestTimeReporter < Minitest::Reporters::BaseReporter
+      include ::CI::Queue::OutputHelpers
+
+      def initialize(build:, limit: nil, percentile: nil, **options)
+        super(options)
+        @test_time_hash ||= build.fetch
```
esnunes: Why using `||=` here?

ruby/test/minitest/queue/test_time_reporter_test.rb
```diff
@@ -0,0 +1,64 @@
...
+        build = mock_build({'some test': [0.01, 0.03, 0.2]})
+        reporter = TestTimeReporter.new(build: build, limit: 0.02, percentile: 0.5)
+
+        output, err = capture_subprocess_io do
+          reporter.report
+        end
+
+        expected_output = <<~EOS
+          +++ Test Time Report
+          \e[31mDetected 1 test(s) over the desired time limit.\e[0m
```
sander-m: Can't you use the `normalize(...)` method here to decolorize the output?

ruby/lib/minitest/queue/test_time_reporter.rb
```diff
@@ -0,0 +1,69 @@
...
+
+        if offending_tests.empty?
+          msg = "The #{humanized_percentile} of test execution time is within #{limit} milliseconds."
+          puts green(msg)
+          return
+        end
+
+        @success = false
+        puts <<~EOS
+          #{red("Detected #{offending_tests.length} test(s) over the desired time limit.")}
```
sander-m: Nitpick: I know that we don't have Rubocop configured for this project, but using `.size` on arrays is usually preferred over `.length` (with `.length` used for string lengths).

ruby/lib/minitest/queue/test_time_reporter.rb
```diff
@@ -0,0 +1,69 @@
...
+          test_time_hash.each_with_object({}) do |(test_name, durations), offenders|
+            duration = calculate_percentile(durations)
+            next if duration <= limit
+            offenders[test_name] = duration
+          end
+        end
+      end
+
+      def calculate_percentile(array)
+        array.sort[(percentile * array.length).ceil - 1]
```
sander-m: Do we know beforehand how many items exist in the result array? Calculating a percentile value only makes sense when the sample set is large enough, e.g. anything under 3 items is usually not meaningful.

ruby/lib/minitest/queue/runner.rb
```diff
@@ -362,34 +373,34 @@ def parser
...
           end
 
           help = <<~EOS
             Defines how long ci-queue should maximally run in seconds
             Defaults to none.
           EOS
           opts.separator ""
-          opts.on('--max-duration SECONDS', *help) do |max|
-            queue_config.max_duration = Integer(max)
+          opts.on('--max-duration SECONDS', Integer, help) do |max|
```
doodzik: this should be in milliseconds and not seconds as the unit is too big.

ruby/lib/minitest/queue/test_time_reporter.rb
doodzik: It depends on how someone uses the grind command. You can configure how often you want the test to be grinded.

ruby/lib/minitest/queue/runner.rb
jessicaxiejw: This is old code. If we are changing it, we should change it in a separate PR. This will also be a breaking API change for anybody using the gem.

ruby/lib/minitest/queue/test_time_reporter.rb
jessicaxiejw: In core, we run each test 1000 times. This should be enough for percentile.

ruby/test/minitest/queue/test_time_reporter_test.rb
jessicaxiejw: That's nifty. I didn't know about the command.

jessicaxiejw: Thanks for telling me!

ruby/lib/minitest/queue/runner.rb
doodzik: We haven't released the feature yet so I wouldn't be concerned with breaking the API. If you want to tackle it in a separate PR please create an issue and add it to our kanban board. So we can track it :)

doodzik: Oh shooot. Forget what I saw. I commented on the wrong option.

ruby/lib/minitest/queue/test_time_reporter.rb
sander-m: Right. For general use though, I think it would make sense to print a warning if less than 3 items exist in the array. It's generally considered wrong to calculate percentile scores for 1 or 2 items in an array.

doodzik: I think that makes sense.

jessicaxiejw: Since the command is used for grind, it is assumed that the number is big enough that the grind will give you an accurate result. What I am trying to say is that if there is a minimum number for percentile, we should have a minimum number for grind.

jessicaxiejw: But the present of a minimum number of grind will warrant an accurate percentile.

sander-m: That's why I was asking "Do we know beforehand how many items exist in the result array?". If we know that with certainty (i.e., the check is already somewhere else), I'm fine with that too.

doodzik: It might be ok to discard a grind if it is less than 50 runs. That is probably the easiest way to do that.

sander-m: > Since the command is used for grind

If the code allows for other use cases (not tied to grind), it should independently do those checks and output some warnings.

esnunes: In case in the future we face redis issues to store this data you can refactor to use a bucket approach: https://stackoverflow.com/a/1249003/1542719

Anyway, we should monitor the redis instance as this will probably increase the redis calls

PR #118, opened 2019-12-11
save test metrics

doodzik:
This PR adds the ability to output data on test results, which is we use for data analysis.
Doing this through json instead of the junit xml gives us an easier time using it and also more flexibility.

ref https://github.com/Shopify/test-infra/issues/100

### DIFF
--- a/ruby/test/integration/minitest_redis_test.rb
+++ b/ruby/test/integration/minitest_redis_test.rb
@@ -9,6 +9,8 @@ module Integration
     def setup
       @junit_path = File.expand_path('../../fixtures/log/junit.xml', __FILE__)
       File.delete(@junit_path) if File.exist?(@junit_path)
+      @test_data_path = File.expand_path('../../fixtures/log/test_data.json', __FILE__)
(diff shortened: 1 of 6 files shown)


ruby/test/integration/minitest_redis_test.rb
```diff
@@ -256,6 +258,78 @@ def test_down_redis
...
+      end
+      assert_empty err
+      output = normalize(out.lines.last.strip)
+      assert_equal 'Ran 9 tests, 6 assertions, 1 failures, 1 errors, 1 skips, 2 requeues in X.XXs', output
+
+      content = File.read(@test_data_path)
+      failure = JSON.parse(content, symbolize_names: true)
+                    .sort_by { |h| "#{h[:test_id]}##{h[:test_index]}" }
+
+      assert_equal failure[0][:namespace], 'foo'
```
sander-m:
```suggestion
      assert_equal 'foo', failure[0][:namespace]
```

doodzik: changed

PR #119, opened 2020-01-02
Verbose mode

lastgabs:
Adding logs to ci-queue. Closing this PR as it shouldn't be merged, but here it is in case anyone ever needs it.


edward:
Looks like the Spy repo is now failing because of a [dependency in its Gemfile](https://github.com/Shopify/spy/blob/master/Gemfile#L126) on this branch.

Is it okay for spy to depend on ci-queue/master instead?

casperisfine:
I released 0.17.0 just a couple hours ago. Spy can use that.

edward:
Thanks @casperisfine

PR #121, opened 2020-02-04
Add option to disable TestTimeReporter

jessicaxiejw:
Related: https://github.com/Shopify/test-infra/issues/183

### DIFF
--- a/ruby/lib/minitest/queue/runner.rb
+++ b/ruby/lib/minitest/queue/runner.rb
@@ -98,7 +98,7 @@ module Minitest
           TestDataReporter.new(namespace: queue_config&.namespace),
         ]
 
-        if queue_config.track_execution_duration
+        if queue_config.track_test_duration
           test_time_record = CI::Queue::Redis::TestTimeRecord.new(queue_url, queue_config)
           reporters << TestTimeRecorder.new(build: test_time_record)
         end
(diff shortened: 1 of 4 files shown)


ruby/test/integration/grind_redis_test.rb
```diff
@@ -256,6 +259,7 @@ def test_grind_max_test_duration_failing
           '--worker', '1',
           '--timeout', '5',
           '--max-test-duration', '0.00001',
+          '--track-execution-duration',
```
sander-m: Since the other flag is called `max-test-duration`, wouldn't it be more consistent to call this flag `track-test-duration`?

ruby/lib/minitest/queue/runner.rb
```diff
@@ -205,15 +209,21 @@ def report_grind_command
...
         grind_reporter.report
 
-        test_time_record = CI::Queue::Redis::TestTimeRecord.new(queue_url, queue_config)
-        test_time_reporter = Minitest::Queue::TestTimeReporter.new(
-          build: test_time_record,
-          limit: queue_config.max_test_duration,
-          percentile: queue_config.max_test_duration_percentile,
-        )
-        test_time_reporter.report
+        test_time_reporter_success = if queue_config.track_test_duration
```
doodzik: Can you add a check if `queue_config.max_test_duration_percentile` but not the track test duration. This can help during a potential misconfiguration.

jessicaxiejw: There is a default value on max_test_duration_percentile, which is 0.5. I thought about adding a check, but because of the default percentile, the resulting build will check if the average time is less than 0s. It will be a really easy to debug bug.

Issue #131, opened 2020-02-28
Expire old test runs


tomchipchase:
Looking at our graphs of redis memory usage, it looks like this gem only removes keys when it hits the memory limit of redis.

This is causing us some problems, as the redis SaaS we are using has a total memory limit for all our instances, and our db for the ci-test-queue using all the memory it can use means other databases have less memory available than they need to operate properly.

I thought I would investigate whether there was a way to clear out old test runs from the ci-queue. 
If this isn't currently possible, I'd be happy to put in the work to make it possible. My thought would be to add an optional ttl to the redis commands storing data, but I'd be interested in hearing alternative approaches.

We're using the rubygem version 0.16.0.

casperisfine:
> it looks like this gem only removes keys when it hits the memory limit of redis.

Yeah, it's a lack of documentation. `ci-queue` currently expects the Redis server to have an eviction policy set to `allkeys-lru`.

So it's not the gem that clears the redis, it relies on expiration.

> I thought I would investigate whether there was a way to clear out old test runs from the ci-queue.

In theory we could clear the Redis state when reporting a successful run, however:

  - You can't clear failed run, because the whole point is that workers are idempotent. If you retry a worker it re-use the existing state.
  - It would be quite brittle.

> My thought would be to add an optional ttl to the redis commands storing data

That's something I'd be positive about. Beside the lacking documentation, requiring users to configure their Redis with `maxmemory-policy allkeys-lru` makes the gem harder to use.

PR #132, opened 2020-03-06
Add a ttl for redis keys

tomchipchase:
Following on from #131 

This adds a command line argument for rspec for an expiry time, and sets the expiry for keys in redis when they are set.

I didn't look at the minitest files, but I can take a look if you're happy with this approach.

### DIFF
--- a/ruby/lib/ci/queue/configuration.rb
+++ b/ruby/lib/ci/queue/configuration.rb
@@ -4,7 +4,7 @@ module CI
     class Configuration
       attr_accessor :timeout, :build_id, :worker_id, :max_requeues, :grind_count, :failure_file
       attr_accessor :requeue_tolerance, :namespace, :seed, :failing_test, :statsd_endpoint
-      attr_accessor :max_test_duration, :max_test_duration_percentile, :track_test_duration
+      attr_accessor :max_test_duration, :max_test_duration_percentile, :track_test_duration, :report_expires_in
       attr_reader :circuit_breakers
 
       class << self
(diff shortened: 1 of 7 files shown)


tomchipchase:
> I'd rather have a sane default (thinking 8 hours) rather than check for the option presence everywhere.

I wanted to leave the existing behaviour if it wasn't present, but I can add the default if you think thats better.

casperisfine:
I understand the reasoning, but I don't think it's worth it. 

Either the redis is configured with allkeys-lru and then the `expire` don't really matter, or wether it's not, and the `expire` is desired.

tomchipchase:
Hey, in case you didn't see it, I made the requested changes.

ruby/lib/ci/queue/configuration.rb
```diff
@@ -30,7 +30,7 @@ def initialize(
         timeout: 30, build_id: nil, worker_id: nil, max_requeues: 0, requeue_tolerance: 0,
         namespace: nil, seed: nil, flaky_tests: [], statsd_endpoint: nil, max_consecutive_failures: nil,
         grind_count: nil, max_duration: nil, failure_file: nil, max_test_duration: nil,
-        max_test_duration_percentile: 0.5, track_test_duration: false
+        max_test_duration_percentile: 0.5, track_test_duration: false, report_expires_in: 28_800
```
casperisfine: Nitpick: for readability `8 * 60 * 60` would be prefereable.

kim76:
Hi @tomchipchase ! Did you want to continue with this PR or should we close it?

tomchipchase:
@kim76 I'm not using this gem any more - and I think we found a different solution when I was - so feel free to close it. It might be useful for someone though?

PR #134, opened 2020-04-18
add max test failures feature

doodzik:
This PR adds a feature to CI-Queue that allows us to short circuit the test workers and summary step. If a build encounters many test failures, then the probability is hight that most of those tests originate from a similar failure source. Short-circuiting test runs in case we have many test failures that allow us to give developers faster feedback and reduce the load on our CI system.

integration build: https://buildkite.com/shopify/shopify-selective-tests/builds/139838

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -14,10 +14,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - name: Set up Ruby 2.6.5
+      - name: Set up Ruby 2.6.x
         uses: actions/setup-ruby@v1
         with:
-          ruby-version: 2.6.5
+          ruby-version: 2.6.x
       - name: Run Ruby tests
         run: |
           bin/before-install
--- a/ruby/lib/ci/queue/circuit_breaker.rb
+++ b/ruby/lib/ci/queue/circuit_breaker.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 module CI
   module Queue
-    class CircuitBreaker
+    module CircuitBreaker
       module Disabled
         extend self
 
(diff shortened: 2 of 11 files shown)


ruby/lib/ci/queue/configuration.rb
```diff
@@ -30,30 +31,31 @@ def initialize(
         timeout: 30, build_id: nil, worker_id: nil, max_requeues: 0, requeue_tolerance: 0,
         namespace: nil, seed: nil, flaky_tests: [], statsd_endpoint: nil, max_consecutive_failures: nil,
         grind_count: nil, max_duration: nil, failure_file: nil, max_test_duration: nil,
-        max_test_duration_percentile: 0.5, track_test_duration: false
+        max_test_duration_percentile: 0.5, track_test_duration: false, max_visible_failures: nil
       )
-        @circuit_breakers = [CircuitBreaker::Disabled]
```
doodzik, line 35: I ordered the list alphabetically

ruby/lib/ci/queue/circuit_breaker.rb
```diff
@@ -50,25 +50,27 @@ def current_timestamp
         end
       end
 
-      def initialize(max_consecutive_failures:)
-        @max = max_consecutive_failures
-        @consecutive_failures = 0
-      end
+      class MaxConsecutiveFailures
```
doodzik, line 53: The MaxConsecutiveFailures circuit breaker was defined as the CircuitBreaker class. This cleans it up and puts it into the correct class.

.github/workflows/tests.yml
```diff
@@ -14,10 +14,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - name: Set up Ruby 2.6.5
```
doodzik, line 17: CI was broken

ruby/lib/ci/queue/redis/build_record.rb
```diff
@@ -62,7 +72,10 @@ def fetch_stats(stat_names)
           counts = redis.pipelined do
             stat_names.each { |c| redis.hvals(key(c)) }
           end
-          stat_names.zip(counts.map { |values| values.map(&:to_f).inject(:+).to_f }).to_h
```
doodzik, line 65: Rubocop complained about this line of code being too long.

.github/workflows/tests.yml
jessicaxiejw: CI is fixed now, right? Should we revert this

doodzik: GitHub actions doesn't support the 2.6.5 version of ruby. So I had to fix the tests.yml file to make CI pass. Nothing to revert here.

ruby/lib/ci/queue/static.rb
```diff
@@ -64,9 +64,24 @@ def acknowledge(test)
...
+      end
+
+      def test_failures
+        @test_failures ||= 0
+      end
+
+      def max_test_failures_reached?
+        return false if config.max_test_failures.nil?
+
+        test_failures >= config.max_test_failures
```
jessicaxiejw, line 78: I think there is a difference of meaning between "test_failures" and "failed_tests". I think the latter's meaning is more acute in the sense that it is counting the tests that failed, not counting the requeues.

jessicaxiejw:
Should we put a message about the tests ended early because it has detected too many failures?

PR #137, opened 2020-04-24
Add CI::Queue.requeueable to define wether a failure should be requeued

casperisfine:
For context I'm using https://github.com/Shopify/deprecation_toolkit to fix Ruby 2.7 keyword argument warnings. The problem is that Ruby only fire these warnings once, so they have a significant false nagative change if they are retried.

But even beyond this, I think it's an useful API, we could probably have other uses for it.

Thoughts?

### DIFF
--- a/ruby/lib/ci/queue.rb
+++ b/ruby/lib/ci/queue.rb
@@ -18,10 +18,14 @@ module CI
   module Queue
     extend self
 
-    attr_accessor :shuffler
+    attr_accessor :shuffler, :requeueable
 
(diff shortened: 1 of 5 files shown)


casperisfine:
Example usage:

```ruby
if defined? CI::Queue
  CI::Queue.requeueable = -> (result) do
    !result.failures.any? do |failure|
      failure.error.is_a?(DeprecationToolkit::Behaviors::DeprecationIntroduced)
    end
  end
end
```

PR #141, opened 2020-05-10
Clean up some old backwards compatibility code that we can get rid off

wvanbergen:
By bumping the required minitest version to 5.14, we can get rid of a bunch of old cruft:
- We can assume `Minitest::Result` exists.
- It's spelled `Minitest`, not `MiniTest`

~Depends on #140.~

### DIFF
--- a/ruby/ci-queue.gemspec
+++ b/ruby/ci-queue.gemspec
@@ -31,7 +31,7 @@ Gem::Specification.new do |spec|
 
   spec.add_development_dependency 'bundler'
   spec.add_development_dependency 'rake'
-  spec.add_development_dependency 'minitest', ENV.fetch('MINITEST_VERSION', '~> 5.11')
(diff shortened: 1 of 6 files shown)


DazWorrall:
5.14 is the latest release and relatively new (Jan this year). services-db show most of our apps with minitest are 5.11+ - any idea if they will be able to upgrade relatively painlessly?

wvanbergen:
Looks like we only require version 5.12, in which `Minitest::Result` was introduced. I'll update.

casperisfine:
The downside to this is that we don't set a hard dependency on minitest, so users might still use old versions.

wvanbergen:
> The downside to this is that we don't set a hard dependency on minitest, so users might still use old versions.

Hmmm, forgot about that. I'll think about it. I'll revert for now.

wvanbergen:
Reverted, and re-opened in https://github.com/Shopify/ci-queue/pull/143.

PR #143, opened 2020-05-11
Revert "Revert "Clean up some old backwards compatibility code that we can get rid off""

wvanbergen:
Reverts Shopify/ci-queue#142

We reverted #141 because we don't have a runtime dependency on minitest, only a development dependency which won't get enforced when people install it in their project.


casperisfine:
Also I just checked, even Rails 6 depends on `~> 5.1`, so there might be old versions being still used.

One solution could be to split in multiple gems. `ci-queue-core`, `ci-queue-minitest`, `ci-queue-rspec`. Not sure if it's much better though.

PR #144, opened 2020-05-13
 JUnitXML reporter: add run-command property to test case tags

wvanbergen:
When generating JUnitXML, include a `run-command` attribute in every `<testcase>` tag. Readers of the file can use this to tell users how they can re-run tests to diagnose issues.

Because test runners differ between environments, we allow people to register a `run_command_formatter` that returns the run command. We ship it with a basic one that runs tests using `bundle exec ruby -Ilib:test<testfile> -n `<testname>`, and a different formatter for when the Rails test runner is found.

### DIFF
--- a/ruby/lib/minitest/queue/junit_reporter.rb
+++ b/ruby/lib/minitest/queue/junit_reporter.rb
@@ -79,7 +79,8 @@ module Minitest
             'classname' => suite,
             'assertions' => test.assertions,
             'time' => test.time,
-            'flaky_test' => test.flaked?
+            'flaky_test' => test.flaked?,
+            'run-command' => Minitest.run_command_for_runnable(test),
           }
           attributes['lineno'] = lineno if lineno != -1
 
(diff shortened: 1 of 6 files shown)


ruby/lib/minitest/queue/junit_reporter.rb
```diff
@@ -79,7 +79,8 @@ def add_tests_to(testsuites, suite, tests)
             'classname' => suite,
             'assertions' => test.assertions,
             'time' => test.time,
-            'flaky_test' => test.flaked?
+            'flaky_test' => test.flaked?,
+            'run-command' => Minitest.run_command_for_runnable(test),
```
DazWorrall, line 83: Mixing hyphens and underscores on the keys here, not a huge thing just thought I’d call it out.

wvanbergen: `flaky_test` was the mistake here: underscores are not really common in XML compared to dashes.

PR #147, opened 2020-05-20
Properly remove project root from paths

### LINKED ISSUE #146: Project root gets greedily stripped from failure backtraces in JUnit Reporter
Looks like this `gsub` is removing all the project roots to make paths relative, but causes issues when the project root path/name is also within a nested subdirectory:

https://github.com/Shopify/ci-queue/blob/9396660c6e8c6170730b0c6fb184c1513a0981b9/ruby/lib/minitest/queue/junit_reporter.rb#L104

Take this example path in a backtrace: `components/fooservices/foo/bar.rb`. It should be `components/foo/app/services/foo/bar.rb`. The `app/` project root is being removed from within paths when it should just be at the start.

We probably have to iterate over each line in `backtrace` and only remove it from the start of the line? Or use the relative path helpers.

cc @wvanbergen

wvanbergen:
Fixes #146.

We were too eagerly removing the project root from paths we encounter in assertion messages. Specifically, because our application root is `/app/` on CI, we swould be stripping out path from with paths as well (e.g. `/app/components/app/test/blah.rb` would have the second `/app` stripped).

This fix uses a regular expression to make sure we only remove the project root it from the beginning of stacktrace lines, where they are always at the same place on the line.

### DIFF
--- a/ruby/lib/minitest/queue/junit_reporter.rb
+++ b/ruby/lib/minitest/queue/junit_reporter.rb
@@ -96,13 +96,16 @@ module Minitest
         message.lines.first.chomp.gsub(/\e\[[^m]+m/, '')
       end
 
+      def project_root_path_matcher
+        @project_root_path_matcher ||= %r{(?<=\s)#{Regexp.escape(Minitest::Queue.project_root)}/}
+      end
+
       def message_for(test)
         suite = test.klass
         name = test.name
         error = test.failure
 
-        message_with_relative_paths = error.message.gsub("#{Minitest::Queue.project_root}/", '')
-
+        message_with_relative_paths = error.message.gsub(project_root_path_matcher, '')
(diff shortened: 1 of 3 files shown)


ruby/lib/minitest/queue/junit_reporter.rb
```diff
@@ -101,8 +101,7 @@ def message_for(test)
         name = test.name
         error = test.failure
 
-        message_with_relative_paths = error.message.gsub("#{Minitest::Queue.project_root}/", '')
-
+        message_with_relative_paths = error.message.gsub(%r{^(\s+)#{Regexp.escape(Minitest::Queue.project_root)}/}, '\1')
```
casperisfine, line 104: Could we memoize that regexp? Compiling regexp is suprisingly expensive.

PR #151, opened 2020-06-03
Remove dependency on minitest-reporters

wvanbergen:
This is more of a proof of concept than anything else at this point. 

Because `minitest-reporters` does not play well with the minitest plugin ecosystem, it can cause issues. E.g. if you are using `Minitest::Reporters.use!` in a `test_helper.rb` it will silently get rid of any reporter that plugins may want to register in their plugin hook. Specifically, people may overwrite the reporters that are necessary for ci-queue to work properly. This can cause builds that are supposed to be red to show up green, which has happened in two of our repositories.

I think we have two ways to deal with this: either embrace minitest-reporters with everything we do, or stop using it and use minitest plugins instead. There's pros and cons for both.

In this PR I am exploring the second option: not using minitest-reporters, but use a minitest plugin instead. The implementation is not complete (test failures are related to not using minitest-reporters's backtrace silencer). Also, it doesn't really defend against people using `Minitest::Reporters.use!` and overwrite these reporters, so we would have to do something about that.

At this point, I am open to feedback about this approach, and whether this makes sense to go down this path. Or alternatively, why we should go down the `Minitest::Reporters` path.

### DIFF
--- a/ruby/ci-queue.gemspec
+++ b/ruby/ci-queue.gemspec
@@ -35,7 +35,6 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'rspec', '~> 3.7.0'
   spec.add_development_dependency 'redis', '~> 3.3'
   spec.add_development_dependency 'simplecov', '~> 0.12'
-  spec.add_development_dependency 'minitest-reporters', '~> 1.1'
 
   spec.add_development_dependency 'snappy'
   spec.add_development_dependency 'msgpack'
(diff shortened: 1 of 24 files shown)


casperisfine:
I don't really have any feedback on this TBH. I used minitest-reporters because at the time it seems to be the easiest way to hook in minitest.

doodzik:
I think it might make sense to switch, because it would make the configuration less error prone. I ran into issue with the particular issues that you highlighted.

PR #153, opened 2020-06-16
Add minitest reporters as a dependency of this project

Edouard-chin:
This is required since  github.com/Shopify/ci-queue/commit/ef11b11d79048a3e951480669ab1d5f60b690b49


casperisfine:
I specifically didn't add it as an explicit dependency because the same gem cover both rspec and minitest so I didn't want to install minitest-reports in rspec projects.

Can you elaborate on why you want this change?

PR #159, opened 2020-08-03
Add error_message to ErrorReporter

ChrisBr:
This will make it easier to navigate the failures of the stabilization builds of the test oversight service.

This is my first PR to ci-queue so 

image

Part of https://github.com/Shopify/test-infra/issues/411

### DIFF
--- a/ruby/lib/minitest/queue/failure_formatter.rb
+++ b/ruby/lib/minitest/queue/failure_formatter.rb
@@ -29,6 +29,7 @@ module Minitest
           test_name: test.name,
           test_suite: test.klass,
           error_class: test.failure.exception.class.name,
+          error_message: test.failure.exception.message,
           output: to_s,
         }
       end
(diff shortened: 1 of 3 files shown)


ruby/lib/minitest/queue/failure_formatter.rb
```diff
@@ -29,6 +29,7 @@ def to_h
           test_name: test.name,
           test_suite: test.klass,
           error_class: test.failure.exception.class.name,
+          error_message: test.failure.exception.message,
```
doodzik, line 32: I would include the backtrace in the error_message as well, since our users would want to know that information as well. You can reuse the code from line 63 in this file for this, just make sure to remove the terminal colours.

doodzik:
Can you bump the version number for the gem as well? In that case Shipit will automatically release a new version after you merge this PR.

ChrisBr:
I discussed this with @esnunes and we decided to reuse the `output` attribute because it holds (almost) identical data. We will parse & remove the terminal color data in test-oversight. The reason is that the error message including backtrace can be quite big and we don't want to store additional (almost) identical data in Redis. @doodzik what do you think?

doodzik:
Sounds good 👍

ChrisBr:
We implemented this in a different way.

PR #162, opened 2020-11-12
[WIP] Run tests in a child process to implement reliable timeouts

casperisfine:
cc @ChrisBr

### DIFF
--- a/ruby/lib/minitest/queue.rb
+++ b/ruby/lib/minitest/queue.rb
@@ -67,12 +67,6 @@ module Minitest
     end
   end
 
-  class TimedOut < UnexpectedError
-    def result_label
-      "TimedOut"
-    end
-  end
-
   module Requeueing
     # Make requeues acts as skips for reporters not aware of the difference.
     def skipped?
@@ -167,12 +161,12 @@ module Minitest
         id <=> other.id
       end
 
-      def build_result(failures)
+      def build_result(failures = [])
         example = @runnable.new(@method_name)
         result = Minitest::Result.new(example.name)
         result.klass      = example.class.name
         result.failures   = failures
-        result.time       = 30 # TODO: fix this
+        result.time       = 30
 
         result.source_location = example.method(example.name).source_location rescue ["unknown", -1]
         result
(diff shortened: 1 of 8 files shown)


ruby/lib/minitest/queue.rb
```diff
@@ -168,6 +181,163 @@ def flaky?
...
 
+    class PipeQueue
+      def initialize(deliminiter: nil)
+        @in, @out = IO.pipe
+        @deliminiter = deliminiter
+      end
+
+      def push(object)
+        payload = Marshal.dump(object)
+        out.write("#{payload}#{deliminiter}")
```
casperisfine, line 192: Rather than a delimiter, it might make more sense to prefix the payload with it's size, e.g:

`[payload.bytesize].pack('N')`.

This way the child can read the header that is a known size, interpret it and then know exactly how much more bytes it needs to read.

casperisfine: To clarify, I mean to to write `<payload-size><payload>`. This way you can keep a single pipe, rather than one for lenght and one for payloads.

ruby/lib/minitest/queue.rb
```diff
@@ -168,6 +181,163 @@ def flaky?
...
+        payload.length
```
casperisfine, line 193: Always use `String#bytesize` for this sorts of thing. In this case the strings should be BINARY, so it shouldn't change anything, but if the strings were to be UTF-8, length returns the numbers of characters which can be less that the actual string number of bytes.

So for clarity `bytesize` is way better (and also sligthly faster).

ruby/lib/minitest/queue/runner.rb
```diff
@@ -489,6 +489,10 @@ def parser
           opts.on('--failing-test TEST_IDENTIFIER') do |identifier|
             queue_config.failing_test = identifier
           end
+
+          opts.on('--execution_type type') do |type|
```
casperisfine, line 493: I don't think we plan to have tons of these, so just `--fork` or `--enable-fork` would do I think.

ruby/lib/minitest/queue.rb
```diff
@@ -168,6 +181,163 @@ def flaky?
       end
     end
 
+    class PipeQueue
+      def initialize(deliminiter: nil)
+        @in, @out = IO.pipe
+        @deliminiter = deliminiter
```
kim76, line 187: sp? `delimiter`

kim76 (commented):
I'm very interested to see what kind of performance impact this will have

PR #165, opened 2021-01-17
Gracefully handle invalid JUnit XML files

ChrisBr:
This adds some rescue to gracefully handle invalid JUnit XML files.

AFAIK we only use the JUnit files for annotation so we can safely just ignore an invalid character.

We already do a [similar thing in shopify-build](https://github.com/Shopify/shopify-build/blob/4b6765680d064b14dc23c48c2f6ccd49d9885365/lib/shopify_build/artifact_export_directory.rb#L107).

https://buildkite.com/shopify/shopify-selective-tests/builds/303594#0717c9a3-c2d8-4031-92b6-34c014a40320/153-154

### DIFF
--- a/ruby/lib/minitest/queue/junit_reporter.rb
+++ b/ruby/lib/minitest/queue/junit_reporter.rb
@@ -7,6 +7,8 @@ require 'fileutils'
 module Minitest
   module Queue
     class JUnitReporter < Minitest::Reporters::BaseReporter
+      include ::CI::Queue::OutputHelpers
+
       def initialize(report_path = 'log/junit.xml', options = {})
         super({})
         @report_path = File.absolute_path(report_path)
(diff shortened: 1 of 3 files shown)


ruby/lib/minitest/queue/junit_reporter.rb
```diff
@@ -76,6 +76,8 @@ def add_tests_to(testsuites, suite, tests)
 
           testcase = testsuite.add_element('testcase', attributes)
           add_xml_message_for(testcase, test) unless test.passed?
+        rescue REXML::ParseException, RuntimeError
+          # noop
```
apostolisstergiou-shopify, line 80: Should we print any message here and let the user know we are handling the parsing exception?

apostolisstergiou-shopify (approved):

ruby/lib/minitest/queue/junit_reporter.rb
casperisfine: I was about to suggest the same. We should log the error.

casperisfine (approved):

PR #169, opened 2021-04-08
Exit 1 when there is an uninitialized constant

apostolisstergiou-shopify:
We are currently exiting with 0 when there is an error such as
uninitialized constant. Buildkite and probably other CI systems will
then consider this status as successful and the suite will only fail
when the aggregate of the executed tests is checked.
With this change the individual steps will also fail.

This also changes the behavior of the before-suite failures which will
now start exiting with 1, see https://github.com/Shopify/ci-queue/pull/73/files.

### DIFF
--- a/ruby/lib/rspec/queue.rb
+++ b/ruby/lib/rspec/queue.rb
@@ -398,7 +398,7 @@ module RSpec
           end
         end
 
-        return 0 if @world.non_example_failure
+        return 1 if @world.non_example_failure
         success ? 0 : @configuration.failure_exit_code
       end
 
--- a/ruby/test/fixtures/uninitialized_constant_suite/spec/dummy_spec.rb
+++ b/ruby/test/fixtures/uninitialized_constant_suite/spec/dummy_spec.rb
@@ -1,0 +1,6 @@
+# frozen_string_literal: true
+RSpec.describe NonExistingObject do
+  it "won't run" do
+    expect(1 + 1).to be == 2
+  end
+end
--- a/ruby/test/fixtures/uninitialized_constant_suite/.rspec
+++ b/ruby/test/fixtures/uninitialized_constant_suite/.rspec
@@ -1,0 +1,1 @@
+--require spec_helper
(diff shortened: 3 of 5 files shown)


DazWorrall (approved):
World#non_example_failure: https://www.rubydoc.info/github/rspec/rspec-core/RSpec%2FCore%2FWorld:non_example_failure

> Used internally to signal that a failure outside of an example has occurred, and that therefore the exit status should indicate the run failed.

So our use of zero looks like an omission to me.

ruby/lib/rspec/queue.rb
```diff
@@ -398,7 +398,7 @@ def run_specs(example_groups)
           end
         end
 
-        return 0 if @world.non_example_failure
+        return 1 if @world.non_example_failure
```
DazWorrall, line 401:
```suggestion
        return @configuration.error_exit_code || @configuration.failure_exit_code if @world.non_example_failure
```

.. would mirror what [rspec itself does](https://github.com/rspec/rspec-core/blob/ea8554afd1a2b63677c6593059fa8f2476181deb/lib/rspec/core/runner.rb#L191), not sure if we want that though.

casperisfine:
> So our use of zero looks like an omission to me.

No, I believe the reasoning was that it would likely be some kind of boot problem (e.g. faulty service or whatever), and that the `--report` step is there to report the actual status. Workers are meant to only fail if they actually ran a failing test, any other failure condition is supposed to `exit 0`.

Of course this was at a time where our CI didn't really have soft failures. So I'm open to reconsider this, but I'm pretty sure that wasn't a mistake.

ruby/test/fixtures/uninitialized_constant_suite/spec/dummy_spec.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+RSpec.describe NonExistingObject do
+  it "won't run" do
+    expect(1 + 1).to be == 2
+  end
+end
```
kim76, line 6: I find this confusing 😕  (Figured it out later when reading the test)
```suggestion
RSpec.describe NonExistingObject do
  it "sample method on  uninitialized constant to test error output" do
    expect(1 + 1).to be == 2
  end
end
```

PR #170, opened 2021-09-13
Fix Python workflow

richardnuno:
Running the python-tests workflow against Python 3.8 was causing pylint issues.

This temporarily skips running the tests against 3.8 as well as linting for all versions since 3.8 is used in the workflow.

More recent versions of Python should be supported but this unblocks releases for now.

### DIFF
--- a/python/railgun.yml
+++ b/python/railgun.yml
@@ -1,0 +1,13 @@
+# This file is for Shopify employees development environment.
+# If you are an external contributor you don't have to bother with it.
+name: ci-queue
+
+vm:
+  image: /opt/dev/misc/railgun-images/default
+  ip_address: 192.168.64.245
+  memory: 1G
+  cores: 2
+volumes:
+  root: 512M
+services:
+   - redis
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -40,11 +40,11 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - name: Install Python 2.7, 3.6, 3.7, 3.8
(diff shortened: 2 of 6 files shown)


python/railgun.yml
```diff
@@ -0,0 +1,13 @@
...
+
+vm:
+  image: /opt/dev/misc/railgun-images/default
+  ip_address: 192.168.64.245
+  memory: 1G
+  cores: 2
+volumes:
+  root: 512M
+services:
+   - redis
```
kim76, line 13: why was this change needed?

kim76 (commented):
👏🏼

python/railgun.yml
richardnuno: I was getting an error (`railgun.yml must exist in the same directory as dev.yml`) when running the tests. Based the change on this [commit](https://github.com/Shopify/ci-queue/commit/c998fc3a59fa2fd4909434b82837873e20e31645) that someone had started to also fix the python workflow issue. It's the same as the ruby [file](https://github.com/Shopify/ci-queue/blob/567f54f53fb1e4d65943b28f4742c2d566df4d66/ruby/railgun.yml) so hopefully these settings should be good.

casperisfine:
Not particularly no. I mean this just fixes the CI. Not much point cutting a release.

kim76 (approved):
The [current version is 0.20.9](https://github.com/Shopify/ci-queue/blob/567f54f53fb1e4d65943b28f4742c2d566df4d66/ruby/lib/ci/queue/version.rb#L5) and hasn't been tagged or released so it seems fine to not have to bump it.

PR #174, opened 2021-09-23
Don't crash on missing result

ChrisBr:
I saw these errors on a branch. It doesn't seem to happen all the time though.

image

### DIFF
--- a/ruby/lib/ci/queue/version.rb
+++ b/ruby/lib/ci/queue/version.rb
@@ -2,7 +2,7 @@
 
 module CI
   module Queue
-    VERSION = '0.21.0'
+    VERSION = '0.21.1'
     DEV_SCRIPTS_ROOT = ::File.expand_path('../../../../../redis', __FILE__)
     RELEASE_SCRIPTS_ROOT = ::File.expand_path('../redis', __FILE__)
   end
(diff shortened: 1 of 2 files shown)


casperisfine:
Hum, I'd feel more confortable if we understood what's going on. This PR seem to just put the error under the rug, no?

ChrisBr:
Seems not to be a huge problem but it's anyway a good idea to have this early return.

casperisfine (approved):
Nevermind, I didn't notice we were in `ensure`.

ChrisBr:
> Hum, I'd feel more confortable if we understood what's going on. This PR seem to just put the error under the rug, no?

Yeah so it was my mistake because I patched the `run` method and did not return the result. 

```ruby
def run(*)
  super

  puts "GC stats...."
end
```

But I think it's still a good idea to have an early return here, no? Can it be possible the yield does not return a result here?

richardnuno (approved):
Avoiding the return in the ensure seems like a good suggestion but looks good.

kim76:
@ChrisBr did you want to merge this or close it?

ChrisBr:
Richard fixed it in https://github.com/Shopify/ci-queue/pull/178

PR #178, opened 2021-10-20
Fix NoMethodError when result is not defined

richardnuno:
When the ensure block is called and the result isn't defined, a NoMethodError is thrown.


kingbbello (approved):

ChrisBr:
> When the ensure block is called and the result isn't defined, a NoMethodError is thrown.

When would that happen?

I only ever had this happen when I had a bug in the code, see https://github.com/Shopify/ci-queue/pull/174.

PR #180, opened 2021-12-10
Implement a release command

casperisfine:
If a worker hard crash (e.g. OOMKilled), other workers will have to wait for its reversed test to reach the timeout before they pick it up.

In many cases we could detect these crashes and timeout the dead worker lease immediately.

e.g.:

```bash
minitest-queue run ....
if [ $? -gte 128 ]; # worker was signaled and might have crashed
  minitest-queue release
fi  
```

We could even have the `run` command fork a worker and monitor it to automatically release the reserved tests if it crashed.

cc @DazWorrall @Shopify/test-infra (I'll test this a bit more but opinions already welcome).


casperisfine:
I tested it on our CI and it behaved as expected. Please review.

ruby/Gemfile
```diff
@@ -2,5 +2,5 @@ source 'https://rubygems.org'
 
 # Specify your gem's dependencies in ci-queue.gemspec
 gemspec
-
+gem 'byebug'
```
paarthmadan, line 5: Accidental? If not, perhaps we introduce a development scope?

redis/release.lua
```diff
@@ -0,0 +1,15 @@
...
+
+-- owned_tests = {"SomeTest", "worker:1", "SomeOtherTest", "worker:2", ...}
+local owned_tests = redis.call('hgetall', owners_key)
+for index, owner_or_test in ipairs(owned_tests) do
+  if owner_or_test == worker_queue_key then -- If we owned a test
+    local test = owned_tests[index - 1]
+    redis.call('zadd', zset_key, "0", test) -- We expire the lease immediately
+    return nil
+  end
+end
```
paarthmadan, line 13: Nit: Is it worth iterating by step of 2 and only performing the equality check on the worker queue keys instead? Not sure if the performance implications are worth it, but would reduce # of elementary operations by half.

ruby/test/ci/queue/redis_test.rb
```diff
@@ -118,6 +118,24 @@ def test_timed_out_test_are_picked_up_by_other_workers
     assert_equal [], populate(second_queue.retry_queue).to_a.sort
   end
 
+  def test_release_immeditely_timeout_the_lease
```
paarthmadan, line 121: Nit:
```suggestion
  def test_release_immediately_timeout_the_lease
```

ruby/Gemfile
casperisfine: I just needed it to debug something, but yeah no big reason to include it in this PR.

redis/release.lua
casperisfine: I wanted to do that but couldn't figure out how to do it.

If you know, I can update.

DazWorrall (approved):
LGTM 👏

casperisfine:
Merging now. If you have idea on how to better iterate HGETALL, let me know.

redis/release.lua
paarthmadan: I'm not well-versed in Lua, so after digging into it I realized it is harder than I initially trivialized it to be.

That said, we might be able to query Redis for the length of the hash first, and use that to index into the response of `hgetall`.

Something like:

```lua
-- Can perform this in a transaction (i.e MULTI, EXEC) to guarantee atomicity so len of hash doesn't change
local len = redis.call('hlen', owners_key)
local owned_tests = redis.call('hgetall', owners_key)

for i=1,len do
  tuple_index = i * 2
  local owner = owned_tests[tuple_index + 1]
  if owner == worker_queue_key then
    local test = owned_tests[tuple_index]
    redis.call('zadd', zset_key, "0", test)
    return nil
  end
end
```

Not sure if the extra call to Redis is worth it. If you think this looks suitable, I can quickly throw a PR up

PR #184, opened 2022-05-30
Add support for a custom TTL of Redis keys for Minitest, defaulting to 8 hours

### LINKED ISSUE #131: Expire old test runs

DougEdey:
Fixes #131 

This recreates https://github.com/Shopify/ci-queue/pull/132 but I would like to get it in.

I have not added this to Python (because I don't use it/not a Python developer)

I've also added documentation about the Redis eviction policy, because that bit us in the backside :)

I found a few places where I think the original PR didn't cover all the keys, so this adds those.

To test, I used a single local test in another app:

```
bundle exec minitest-queue --queue redis://localhost run -Itest test/some_test.rb --namespace dougttl3
```

Then I read the namespace keys back in redis and put the TTL on the output to confirm everything has one set:

```
> redis.keys("build:dougttl3:*").each { |key| puts "#{key}: #{redis.ttl(key)}" }

build:dougttl3::worker::queue: 28583
build:dougttl3::processed: 28583
build:dougttl3::total: 28581
build:dougttl3::master-status: 28581
build:dougttl3::workers: 28581
```

This test doesn't set the worker id or build id but those would be set in CI, if I re-test with the build and worker set manually

```
> redis.keys("build:dougttl3:1:*").each { |key| puts "#{key}: #{redis.ttl(key)}" }

build:dougttl3:1:workers: 28750
build:dougttl3:1:total: 28750
build:dougttl3:1:processed: 28751
build:dougttl3:1:master-status: 28750
build:dougttl3:1:worker:1:queue: 28751
```

### DIFF
--- a/ruby/lib/ci/queue/configuration.rb
+++ b/ruby/lib/ci/queue/configuration.rb
@@ -18,7 +18,7 @@ module CI
             seed: env['CIRCLE_SHA1'] || env['BUILDKITE_COMMIT'] || env['TRAVIS_COMMIT'] || env['HEROKU_TEST_RUN_COMMIT_VERSION'] || env['SEMAPHORE_GIT_SHA'],
             flaky_tests: load_flaky_tests(env['CI_QUEUE_FLAKY_TESTS']),
             statsd_endpoint: env['CI_QUEUE_STATSD_ADDR'],
-            redis_ttl: env['CI_QUEUE_REDIS_TTL'],
+            redis_ttl: env['CI_QUEUE_REDIS_TTL']&.to_i ||  8 * 60 * 60,
           )
         end
 
--- a/ruby/test/ci/queue/configuration_test.rb
+++ b/ruby/test/ci/queue/configuration_test.rb
@@ -65,6 +65,18 @@ module CI::Queue
       assert_equal 'browser:9e08ef3c-d6e6-4a86-91dd-577ce5205b8e', config.build_id
     end
 
+    def test_redis_ttl_defaults
+      config = Configuration.new
+      assert_equal(28_800, config.redis_ttl)
+    end
+
+    def test_redis_ttl_from_env
+      config = Configuration.from_env(
+        "CI_QUEUE_REDIS_TTL" => "14400"
+      )
+      assert_equal(14_400, config.redis_ttl)
+    end
+
     def test_parses_file_correctly
       Tempfile.open('flaky_test_file') do |file|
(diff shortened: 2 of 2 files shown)


ruby/README.md
```diff
@@ -70,3 +70,9 @@ rspec-queue --queue redis://example.com --timeout 600 --report
 #### Limitations
 
 Because of how `ci-queue` executes the examples, `before(:all)` and `after(:all)` hooks are not supported. `rspec-queue` will explicitly reject them.
+
+## Custom Redis Expiry
+
+`ci-queue` expects the Redis server to have an [eviction policy](https://redis.io/docs/manual/eviction/#eviction-policies) of `allkeys-lru`.
+
+You can also use `--report-expires-in` to set a custom expiration time for all CI Queue keys
```
casperisfine, line 78: Would probably be worth setting the default here.

DougEdey: I assume you mean showing?

casperisfine: Yes 😉

casperisfine (approved):
LGTM. I was about to finish my day, so I'll give this another pass tomorrow.

ruby/lib/ci/queue/configuration.rb
```diff
@@ -18,6 +18,7 @@ def from_env(env)
             seed: env['CIRCLE_SHA1'] || env['BUILDKITE_COMMIT'] || env['TRAVIS_COMMIT'] || env['HEROKU_TEST_RUN_COMMIT_VERSION'] || env['SEMAPHORE_GIT_SHA'],
             flaky_tests: load_flaky_tests(env['CI_QUEUE_FLAKY_TESTS']),
             statsd_endpoint: env['CI_QUEUE_STATSD_ADDR'],
+            redis_ttl: env['CI_QUEUE_REDIS_TTL'],
```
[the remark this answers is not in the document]
DougEdey: I also added some tests for this (I was missing a `&.to_i` on the env too

PR #186, opened 2022-06-16
Set seed via new cattr_accessor 'Minitest#seed' explicitly

serioushaircut:
this "fixes" this [issue](https://github.com/minitest/minitest/issues/913)

### DIFF
--- a/ruby/lib/minitest/queue/runner.rb
+++ b/ruby/lib/minitest/queue/runner.rb
@@ -21,6 +21,7 @@ module Minitest
 
       def initialize(argv)
         @queue_config = CI::Queue::Configuration.from_env(ENV)
+        Minitest.seed = @queue_config.seed.to_i
         @command, @argv = parse(argv)
       end
 


ruby/lib/minitest/queue/runner.rb
```diff
@@ -21,6 +21,7 @@ def self.invoke(argv)
 
       def initialize(argv)
         @queue_config = CI::Queue::Configuration.from_env(ENV)
+        Minitest.seed = @queue_config.seed.to_i
```
casperisfine, line 24: Since this was added in the latest version, we want to check if the method exists.

```suggestion
        if Minitest.respond_to?(:seed=)
          Minitest.seed = @queue_config.seed.to_i
        end
```

casperisfine: Also `parse()` takes a `--seed` option, so you want to do that after `parse`.

serioushaircut: both very correct, ty

zenspider:
Lemme know if there's anything I should change on my side to assist. I'm gonna have to put out a bugfix shortly for some mock/kwarg stuff.

casperisfine:
> Could I get a hint on where to build a test to reproduce this behaviour.

Just updating to minitest `5.16` breaks half the test suite. I don't think we need a new test, a dependency update should be enough.

> Lemme know if there's anything I should change on my side to assist.

Thanks, appreciated. What is happening is that since we have our own runner script, we bypass `Minitest.run` and later call `Minitest::Test.runnable_methods`. That's why I think it's on us.

But if you wish to support this maybe `Minitest.seed` could be autogenerated if somehow it's not initialized.

PR #188, opened 2022-07-13
Exit early if all workers die

ChrisBr:
We frequently have the case that all workers after setup but we still wait ~30 min for all workers to finish. We should check if there are any workers left and if not exit early.

https://buildkite.com/shopify/shopify-selective-tests/builds/752110#0181f401-8546-447f-93c1-6863d019b5d9
https://buildkite.com/shopify/shopify-selective-tests/builds/751666#0181f338-af0f-4fb0-9a59-37ce1aa077e8

### DIFF
--- a/ruby/lib/ci/queue/redis/supervisor.rb
+++ b/ruby/lib/ci/queue/redis/supervisor.rb
@@ -21,16 +21,32 @@ module CI
 
           yield if block_given?
 
-          time_left = config.timeout
-          until exhausted? || time_left <= 0 || max_test_failed?
-            sleep 1
+          time_left = config.report_timeout
+          time_left_with_no_workers = config.inactive_workers_timeout
+          until exhausted? || time_left <= 0 || max_test_failed? || time_left_with_no_workers <= 0
             time_left -= 1
+            sleep 1
+
+            if active_workers?
+              time_left_with_no_workers = config.inactive_workers_timeout
+            else
+              time_left_with_no_workers -= 1
+            end
 
             yield if block_given?
           end
+
+          puts "Aborting, it seems all workers died." if time_left_with_no_workers <= 0
           exhausted?
         rescue CI::Queue::Redis::LostMaster
           false
+        end
+
+        private
+
+        def active_workers?
+          # if there are running jobs we assume there are still agents active
+          redis.zrangebyscore(key('running'), Time.now.to_f - config.timeout, "+inf", limit: [0,1]).count > 0
         end
       end
     end
--- a/ruby/lib/minitest/queue/runner.rb
+++ b/ruby/lib/minitest/queue/runner.rb
@@ -368,6 +368,24 @@ module Minitest
           end
 
           help = <<~EOS
+            Specify a timeout after which the report command will fail if not all tests have been processed.
+            Defaults to the value set for --timeout.
+          EOS
+          opts.separator ""
+          opts.on('--report-timeout TIMEOUT', Float, help) do |timeout|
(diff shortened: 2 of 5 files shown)


ruby/test/ci/queue/redis_supervisor_test.rb
```diff
@@ -39,6 +39,18 @@ def test_wait_for_workers
     assert_equal true, workers_done
   end
 
+  def test_wait_for_workers_timeout
+    @supervisor = supervisor(timeout: 10, queue_init_timeout: 0.1)
```
ChrisBr, line 43: I'm not really happy with this test because we can't really distinguish between if we died because of the general timeout or all workers died. I could set the timeout a lot higher but still, the expectation of the test would still be the same. We could add some logging which we assert.

ruby/lib/ci/queue/redis/supervisor.rb
```diff
@@ -22,16 +22,28 @@ def wait_for_workers
...
           end
           exhausted?
         rescue CI::Queue::Redis::LostMaster
           false
         end
+
+        private
+
+        def active_workers?
+          redis.zrangebyscore(key('running'), Time.now.to_f - config.timeout, "+inf").count > 0
```
casperisfine, line 45: Minor micro-opt, you can pass `limit: [0, 1]` to get returned one record at most.

ruby/test/ci/queue/redis_supervisor_test.rb
casperisfine: Yeah, asserting logs sounds OK to me, especially since that's an error condition you'd want to log anyway.

ruby/lib/ci/queue/redis/supervisor.rb
```diff
@@ -21,17 +21,33 @@ def wait_for_workers
 
           yield if block_given?
 
-          time_left = config.timeout
-          until exhausted? || time_left <= 0 || max_test_failed?
-            sleep 1
+          time_left = config.report_timeout
+          time_left_with_no_workers = config.queue_init_timeout
```
ChrisBr, line 25: I wonder if you need to set some minimum default in case somebody set this really low (e.g. to 1 like what we did in Core)?

cc @casperisfine

casperisfine: Yeah, definitely shouldn't be set too low.

ruby/lib/minitest/queue/runner.rb
```diff
@@ -367,6 +367,15 @@ def parser
             queue_config.timeout = timeout
           end
 
+          help = <<~EOS
+            Specify a timeout after which the report command will fail if not all tests have been processed.
```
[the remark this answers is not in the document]
ChrisBr: No, we will use `queue_init_timeout` for this.

https://github.com/Shopify/ci-queue/pull/188/files#diff-9f624fcbe4bfbe499cdf98b7e9a82096670bb53e4d02f98a898548290bda1dd0R24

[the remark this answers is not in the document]
ChrisBr: We basically need three timeouts for the report 

* timeout: how long a test can take. If test execution > timeout it's considered dead
* queue_init_timeout: if there are no active workers, how long will we wait for a worker to come back up
* report_timeout: how long will we wait for all workers to finish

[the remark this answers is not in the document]
casperisfine: Sounds about right.

PR #190, opened 2022-07-20
Release 0.24.0

ChrisBr:
Release 0.23.2

* https://github.com/Shopify/ci-queue/pull/189
* https://github.com/Shopify/ci-queue/pull/188

### DIFF
--- a/ruby/lib/ci/queue/version.rb
+++ b/ruby/lib/ci/queue/version.rb
@@ -2,7 +2,7 @@
 
 module CI
   module Queue
-    VERSION = '0.23.1'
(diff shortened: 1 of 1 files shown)


ruby/lib/ci/queue/version.rb
```diff
@@ -2,7 +2,7 @@
 
 module CI
   module Queue
-    VERSION = '0.23.1'
+    VERSION = '0.23.2'
```
casperisfine, line 5: It's all new features, so `0.24.0` would make more sense.

casperisfine (approved):

PR #199, opened 2023-03-15
Abort when too many test failed

ChrisBr:
We should not assume the reporter will report a non 0 exit code as all tests could be retried.


ruby/test/integration/minitest_redis_test.rb
```diff
@@ -91,8 +91,9 @@ def test_max_test_failed
       assert_equal 'Ran 47 tests, 47 assertions, 3 failures, 0 errors, 0 skips, 44 requeues in X.XXs', output
 
       # Run the reporter
+      exit_code = nil
       out, err = capture_subprocess_io do
-        system(
+        exit_code = system(
```
casperisfine, line 96: You don't use this.

casperisfine (approved):

PR #208, opened 2023-05-02
Ignore Python tests

ChrisBr:
We ignore the Python tests for a while now so let's make it explicit.


ChrisBr:
> I don't ignore Python tests, I just don't maintain the Python side of things, it's handled by another team etc. I wouldn't make this decision for them.

Sounds good. Who is the team responsible for Python?

casperisfine:
Used to be @solackerman, but not sure anymore.

PR #209, opened 2023-05-02
Fix leakbot command instruction in Spin

Bor1s:
Avoid asking for `worker` and `build` while running a leakbot reproducing command in Spin.

Resolves: Shopify/shopify#421914

### DIFF
--- a/ruby/lib/minitest/queue/runner.rb
+++ b/ruby/lib/minitest/queue/runner.rb
@@ -253,8 +253,10 @@ module Minitest
       attr_accessor :queue, :queue_url, :grind_list, :grind_count, :load_paths
 
       def require_worker_id!
-        invalid_usage!("build-id couldn't be inferred from ENV and wasn't set via --build") unless queue_config.build_id
-        invalid_usage!("worker-id couldn't be inferred from ENV and wasn't set via --worker") unless queue_config.worker_id
+        if queue.distributed?
+          invalid_usage!("build-id couldn't be inferred from ENV and wasn't set via --build") unless queue_config.build_id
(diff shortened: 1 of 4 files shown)


casperisfine:
> Trying to understand why Python fails in Github Actions:

The Python side of this project isn't maintained by us, just ignore these failures.

ruby/lib/minitest/queue/runner.rb
```diff
@@ -582,6 +582,14 @@ def retry?
         ENV["BUILDKITE_RETRY_COUNT"].to_i > 0 ||
           ENV["SEMAPHORE_PIPELINE_RERUN"] == "true"
       end
+
+      def reproduce_command
+        if ENV['CI'] != 'true' || queue.is_a?(CI::Queue::File)
```
casperisfine, line 587: I don't think you need this conditional given that it's expected you'd run this repro command locally.

casperisfine:
I think a simpler fix might be to assign a default `worker_id` and `build_id` if the queue is local.

ruby/lib/minitest/queue/runner.rb
ChrisBr: I think the idea was to just not raise in this case here

https://github.com/Shopify/ci-queue/blob/10d895f08eebbe562dcbb75dfbe270a0c7bc565a/ruby/lib/minitest/queue/runner.rb#L255-L258

But just always printing `bundle exec minitest-queue --queue - run --build local --worker local` may work as well.

Bor1s:
Updated the code here: https://github.com/Shopify/ci-queue/pull/209/commits/2d51476232beb5d869e3307bf020d0eba5f20fea
Could you please have another 👁️ @casperisfine @ChrisBr ?

ChrisBr (approved):

PR #211, opened 2023-05-12
Export flaky tests

nikita8:
Export list of flaky tests.

Re-queued jobs are not marked as failed. So, to determine the flakiness I am looking into the re-queue count for the test or check if it failed earlier and passed on job retry.

### DIFF
--- a/ruby/lib/ci/queue/redis/build_record.rb
+++ b/ruby/lib/ci/queue/redis/build_record.rb
@@ -48,9 +48,22 @@ module CI
         end
 
         def record_success(id, stats: nil)
-          redis.pipelined do |pipeline|
+          error_reports_deleted_count, requeued_count, _ = redis.pipelined do |pipeline|
             pipeline.hdel(key('error-reports'), id.dup.force_encoding(Encoding::BINARY))
+            pipeline.hget(key('requeues-count'), id.b)
             record_stats(stats, pipeline: pipeline)
+          end
+          record_flaky(id) if error_reports_deleted_count.to_i > 0 || requeued_count.to_i > 0
+          nil
+        end
+
+        def record_flaky(id, stats: nil)
+          redis.pipelined do |pipeline|
+            pipeline.sadd(
+              key('flaky-reports'),
+              id.b
+            )
+            pipeline.expire(key('flaky-reports'), config.redis_ttl)
           end
           nil
         end
@@ -63,6 +76,10 @@ module CI
 
         def error_reports
           redis.hgetall(key('error-reports'))
+        end
+
+        def flaky_reports
+          redis.smembers(key('flaky-reports'))
         end
 
         def fetch_stats(stat_names)
--- a/ruby/lib/minitest/queue/runner.rb
+++ b/ruby/lib/minitest/queue/runner.rb
@@ -213,6 +213,11 @@ module Minitest
           File.write(queue_config.failure_file, failures)
         end
 
+        if queue_config.export_flaky_tests_file
+          failures = reporter.flaky_reports.to_json
+          File.write(queue_config.export_flaky_tests_file, failures)
+        end
+
         reporter.report
         exit! reporter.success? ? 0 : 1
       end
@@ -479,6 +484,15 @@ module Minitest
           opts.separator ""
           opts.on('--failure-file FILE', help) do |file|
             queue_config.failure_file = file
+          end
+
+          help = <<~EOS
+            Defines a file where flaky tests during the execution are written to in json format.
+            Defaults to disabled.
+          EOS
+          opts.separator ""
+          opts.on('--export-flaky-tests-file FILE', help) do |file|
+            queue_config.export_flaky_tests_file = file
           end
 
           help = <<~EOS
--- a/ruby/lib/ci/queue/configuration.rb
+++ b/ruby/lib/ci/queue/configuration.rb
@@ -2,7 +2,7 @@
 module CI
   module Queue
     class Configuration
-      attr_accessor :timeout, :worker_id, :max_requeues, :grind_count, :failure_file
+      attr_accessor :timeout, :worker_id, :max_requeues, :grind_count, :failure_file, :export_flaky_tests_file
       attr_accessor :requeue_tolerance, :namespace, :failing_test, :statsd_endpoint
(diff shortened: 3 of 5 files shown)


ruby/lib/minitest/queue/build_status_recorder.rb
```diff
@@ -52,6 +52,7 @@ def record(test)
         if (test.failure || test.error?) && !test.skipped?
           build.record_error("#{test.klass}##{test.name}", dump(test), stats: stats)
         else
+          build.record_flaky("#{test.klass}##{test.name}", stats: stats)
```
ChrisBr, line 55: My main concern is that we double the number of Redis calls we do for a successful run.

ChrisBr, line 55: Also might make sense to add this to `record_success` pipeline so we don't need to do two roundtrips.

ruby/lib/minitest/queue/runner.rb
```diff
@@ -213,6 +213,9 @@ def report_command
           File.write(queue_config.failure_file, failures)
         end
 
+        flaky_tests = reporter.flaky_reports.to_json
+        File.write("log/flaky_tests.txt", flaky_tests)
```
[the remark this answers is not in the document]
nikita8: Would it be similar to the failure file?

ruby/lib/ci/queue/redis/build_record.rb
```diff
@@ -48,10 +48,25 @@ def record_error(id, payload, stats: nil)
...
           end
+          record_flaky(id) if deleted_count == 1
+          nil
+        end
+
+        def record_flaky(id, stats: nil)
+          redis.pipelined do |pipeline|
+            pipeline.sadd(
+              key('flaky-reports'),
+              id.dup.force_encoding(Encoding::BINARY)
```
casperisfine, line 63:
```suggestion
              id.b
```

ruby/lib/ci/queue/redis/build_record.rb
```diff
@@ -48,10 +48,25 @@ def record_error(id, payload, stats: nil)
...
+            )
+            pipeline.expire(key('flaky-reports'), config.redis_ttl)
+          end
+          nil
+        rescue => e
+          puts "Error recording flaky: #{e.message}"
```
casperisfine, line 69: Not a fan of this `puts`, as this class isn't supposed to do any UI stuff.

Why do we even expect this to fail?

ruby/lib/minitest/queue/runner.rb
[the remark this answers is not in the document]
ChrisBr: Yes correct!

ruby/lib/ci/queue/redis/build_record.rb
```diff
@@ -48,10 +48,25 @@ def record_error(id, payload, stats: nil)
         end
 
         def record_success(id, stats: nil)
-          redis.pipelined do |pipeline|
+          deleted_count, _ = redis.pipelined do |pipeline|
             pipeline.hdel(key('error-reports'), id.dup.force_encoding(Encoding::BINARY))
             record_stats(stats, pipeline: pipeline)
           end
+          record_flaky(id) if deleted_count == 1
```
ChrisBr, line 55: So we can't read the result of a command inside a Redis pipeline?

ruby/lib/ci/queue/redis/build_record.rb
```diff
@@ -48,10 +48,23 @@ def record_error(id, payload, stats: nil)
         end
 
         def record_success(id, stats: nil)
-          redis.pipelined do |pipeline|
+          deleted_count, requeued_count, _ = redis.pipelined do |pipeline|
```
ChrisBr, line 51: [nit] it's not really `deleted_count`, it's more like checking if the test errored before.

So maybe something like

```suggestion
          prev_error_reported, requeued_count, _ = redis.pipelined do |pipeline|
```

ruby/lib/ci/queue/redis/build_record.rb
```diff
@@ -48,10 +48,23 @@ def record_error(id, payload, stats: nil)
...
             pipeline.hdel(key('error-reports'), id.dup.force_encoding(Encoding::BINARY))
+            pipeline.hget(key('requeues-count'), id.b)
             record_stats(stats, pipeline: pipeline)
           end
+          record_flaky(id) if deleted_count == 1 || requeued_count.to_i > 0
```
ChrisBr, line 56: [nit] `deleted_count == 1` looks more like we deleted 1 element but `1` is, I believe` the exit code here.

So something like this might be more accurate

```suggestion
          SUCCESS_EXIT_CODE = 1
          record_flaky(id) if prev_error_reported == SUCCESS_EXIT_CODE || requeued_count.to_i > 0
```

ruby/lib/minitest/queue/runner.rb
```diff
@@ -481,6 +486,15 @@ def parser
             queue_config.failure_file = file
           end
 
+          help = <<~EOS
+            Defines a file where the flaky tests during the execution are written to in the json format.
```
ChrisBr, line 490:
```suggestion
            Defines a file where flaky tests during the execution are written to in json format.
```

ruby/test/fixtures/log/flaky_tests.txt
```diff
@@ -0,0 +1 @@
+["BTest#test_bar","ATest#test_flaky_fails_retry","ATest#test_flaky","ATest#test_bar"]
```
ChrisBr, line 1: Is this used?

ruby/lib/ci/queue/redis/build_record.rb
nikita8: Based on the [redis doc](https://redis.io/commands/hdel/), it is returning number of fields removed. 

> Integer reply: the number of fields that were removed from the hash, not including specified but non existing fields.

ruby/test/fixtures/log/flaky_tests.txt
nikita8: Oops no, that is the file generated after test execution.

ChrisBr (approved):

PR #213, opened 2023-05-15
Allow specifying files to test with 'dev test'

rwstauner:
The whole suite can take a while, it's nice to be able to run specific files when you're debugging something.

Test all: `dev test`
Test specific files: `dev test test/minitest/reporters/*`


casperisfine:
Note that with minitest you can always run `ruby -Itest test/path/to/test.rb`. Making this PR not strictly necessary. But 👍

ChrisBr (approved):

PR #221, opened 2023-06-14
Export test order log for bisect

ChrisBr:
Export test order log for bisect so we can use it for post processing.


ruby/lib/minitest/queue/runner.rb
```diff
@@ -177,6 +177,8 @@ def bisect_command
           puts
           puts "cat <<EOF |\n#{failing_order.to_a.map(&:id).join("\n")}\nEOF\n#{command.join(' ')}"
           puts
+
+          File.write('log/test_order.log', failing_order.to_a.map(&:id).join("\n"))
```
ChrisBr, line 181: Not sure we want to make this configurable. We already use the hard coded path for the order reporter. Although semantics are slightly different.

PR #225, opened 2023-06-29
Adds support for ordered test runs. 

gustavompo:
This feature is exclusively set to be used to troubleshoot flaky tests that were caught in CI, providing a way to have a rather close environment of a particular build


casperisfine:
This is already supported via `CI_QUEUE_URL=file://path/to/test_order.log`

gustavompo:
> This is already supported via `CI_QUEUE_URL=file://path/to/test_order.log`

Oh ok, totally missed it.

PR #227, opened 2023-07-11
Fix Python tests

sergray:
* Drop end-of-life Python versions bellow 3.7 (Python 3.7 is also end of life, but let's keep it for backward compatibility)
* Use Gitlab CI matrix for installation and test Python versions in parallel (similar to Ruby). Make sure tox runs only selected Python version

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -37,7 +37,7 @@ setuplib.setup(
     packages=['ciqueue', 'ciqueue._pytest'],
     install_requires=[
         'dill>=0.2.7',
-        'pytest>=2.7,<=3.5.0',
+        'pytest>=2.7,<7',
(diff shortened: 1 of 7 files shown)


python/setup.py
```diff
@@ -37,17 +37,17 @@ def get_lua_scripts():
     packages=['ciqueue', 'ciqueue._pytest'],
     install_requires=[
         'dill>=0.2.7',
-        'pytest>=2.7,<=3.5.0',
+        'pytest>=2.7,<7',
```
sergray, line 40: The current https://pypi.org/project/pytest/ is 7.4.0, but upgrading to it is a bigger effort, so make sure we run at least on the latest 6.x

PR #228, opened 2023-07-20
Always write a test order file

ChrisBr:
We always want to write a test order file so we know leak detection run successfully and didn't crash.


davidstosik (approved):
If that means eventually we can make the difference between leakbot crashes and leakbot (successfully) failing to find a reproduction path, then that sounds good to me!

casperisfine (approved):

PR #229, opened 2023-07-27
Ignore broken tests which are not evaluated as part of bisect

ChrisBr:
Ignore broken tests which are not evaluated as part of bisect

image

Kudos to @shioyama for bringing it up!


casperisfine (approved):

shioyama:
Thank you!

davidstosik:
Thank you! ❤️

PR #230, opened 2023-07-28
Add shutdown to Static and File queues, and fix some tests

gmalette:
The `Minitest::Queue::Runner` [traps signals](https://github.com/Shopify/ci-queue/blob/f6159e0c79471d999aa31d40637e7a81f1654731/ruby/lib/minitest/queue/runner.rb#L82-L83) and sends `shutdown!` to the queue. However, `shutdown!` doesn't exist doesn't exist on `Static` and `File` type queues, resulting in a printed exception, and a runaway test runner. This PR adds `shutdown!` to those queues.

It also fixes a test in the `File` type that wasn't run by virtue of being `private`.

### DIFF
--- a/ruby/lib/ci/queue/static.rb
+++ b/ruby/lib/ci/queue/static.rb
@@ -20,6 +20,11 @@ module CI
         @config = config
         @progress = 0
         @total = tests.size
+        @shutdown = false
+      end
+
+      def shutdown!
+        @shutdown = true
       end
 
       def distributed?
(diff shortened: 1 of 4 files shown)


casperisfine (approved):

gmalette:
To create a new release do I just need to bump the version and let shipit handle it, or are there other steps?

ChrisBr:
> To create a new release do I just need to bump the version and let shipit handle it, or are there other steps?

That's right!

PR #238, opened 2023-09-23
Support Redis with TLS

andyundso:
**Background**

We use Semaphore 2 to run our test suites. On a couple of those, we parallise the test execution, but run time between the workers is different, so I wanted to use `ci-queue` to balance it.

Our local cloud provider offers Redis-as-a-service, but only with TLS enabled. The current version of `ci-queue` throws an error when starting the queue url with `rediss://`.

**Implementation**

I first chose to align how the tests are called between Python and Ruby. For both, `REDIS_URL` can be set now to customise the Redis used, otherwise it fallsback to a local connection.

I then added an additional matrix to GitHub Actions that spawns a new set of jobs where both test suites are run against a Redis with TLS enabled. The Redis is setup according to the tutorial [here](https://blog.shahid.codes/setup-redis-with-tls-using-docker).

Afterward, I added new arguments to the different runners and made sure that

1. They use the Redis queue when passing in an URL with `rediss://`.
2. And the test suites passes with the self-signed certificates.

The additions should not make any difference when doing development on the gem or using it in production, as TLS is only activated when passing in `rediss://`.

I'm not really proficient in Python and therefore code is sometimes duplicated, as I was not sure how to properly de-duplicate it. Also, syntax might not be very optimised.


ruby/test/test_helper.rb
```diff
@@ -19,3 +19,5 @@
 Dir[File.expand_path('../support/**/*.rb', __FILE__)].sort.each do |file|
   require file
 end
+
+include RedisHelpers
```
casperisfine, line 23: Please don't include modules in `::Object`, that's a big footgun. 

Include it in the test classes that need it.

casperisfine (commented):
So I don't mind at all supporting `rediss://` URLs, that's trivial and makes sense.

However I'm on the fence about adding more parameters to support private certificates.

I think I'd rather make it easier to programatically instantiate the Queue, e.g.

`minitest-queue --init path/to/script.rb ....`

```ruby
Minitest.queue = CI::Queue::Redis.new(...)
```

This would keep the CLI options low, and give an escape hatch to do all sort of things including custom queue implementations.

casperisfine:
> but I can see that it has little value for actual production scenarios.

I think private certificate for Redis are surprisingly common, I often get issues about it on https://github.com/redis-rb/redis-client and https://github.com/redis/redis-rb. But yeah, annoys me to add lots of code like this to only cover a subset of things. Because very often users need even more OpenSSL tweaking than that.

> Wondering if this something that can be done in Python as well, in case you want to keep the two implementations similar.

For more advanced stuff like this, it's not a big deal if one implementation is lacking.

>  I'm happy to open another PR just containing the necessary changes to support rediss:// URLs.

Please do.

> In case you want to move forward with the init script, I would need bit of guidance how to implement it.

If you don't need it, don't bother. If someone need it one day, they can implement it themselves. I'm OK with `rediss://` not having proper coverage on CI, a mock assertion should be enough.

casperisfine:
Closing this now that #239 was merged.

Issue #156, opened 2020-07-29
Ruby ci-queue requires gems that are not runtime dependencies


RobertMic:
We are getting `LoadError: cannot load such file` errors from the `ci-queue` ruby gem when trying to use it. It looks to me that the gem specifies several `development_dependencies` but uses them outside of development at runtime.

The fix on our end is to require this gems dependencies in our `gemfile`. The gem should properly specify its dependencies so installing it via `bundler` will grab everything it needs.

Here are the errors I've encountered:
```
bundler: failed to load command: minitest-queue (/tmp/bundle/ruby/2.7.0/bin/minitest-queue)
--
  | LoadError: cannot load such file -- redis
  | /tmp/bundle/ruby/2.7.0/gems/ci-queue-0.20.5/lib/ci/queue/redis.rb:3:in `require'
  | /tmp/bundle/ruby/2.7.0/gems/ci-queue-0.20.5/lib/ci/queue/redis.rb:3:in `<top (required)>'
  | /tmp/bundle/ruby/2.7.0/gems/ci-queue-0.20.5/lib/ci/queue.rb:47:in `require'
  | /tmp/bundle/ruby/2.7.0/gems/ci-queue-0.20.5/lib/ci/queue.rb:47:in `from_uri'
  | /tmp/bundle/ruby/2.7.0/gems/ci-queue-0.20.5/lib/minitest/queue/runner.rb:31:in `run!'
  | /tmp/bundle/ruby/2.7.0/gems/ci-queue-0.20.5/lib/minitest/queue/runner.rb:19:in `invoke'
  | /tmp/bundle/ruby/2.7.0/gems/ci-queue-0.20.5/exe/minitest-queue:5:in `<top (required)>'
  | /tmp/bundle/ruby/2.7.0/bin/minitest-queue:23:in `load'
  | /tmp/bundle/ruby/2.7.0/bin/minitest-queue:23:in `<top (required)>'
```
```
bundler: failed to load command: minitest-queue (/tmp/bundle/ruby/2.7.0/bin/minitest-queue)
--
  | LoadError: cannot load such file -- minitest/reporters
  | /tmp/bundle/ruby/2.7.0/gems/ci-queue-0.20.5/lib/minitest/queue.rb:4:in `require'
  | /tmp/bundle/ruby/2.7.0/gems/ci-queue-0.20.5/lib/minitest/queue.rb:4:in `<top (required)>'
[...]

PR #246, opened 2024-01-03
Declare Redis >= 5.0 dependency

pedropb:
Since b403d0c0bfe5d69b9bebad423a4364379496f46e, Redis >= 5 has been a dependency for ci-queue. This commit declares this dependency on the gemspec file.

Relates to #156 .

---

We ran into this issue while troubleshooting a failed dependabot bump on ci-queue. The transitive dependency on `redis` resolved to version 4.8.1 in our app, which satisfied all gems except ci-queue.


pedropb:
It worked for us @ChrisBr . Thanks for the swift response!

PR #247, opened 2024-01-03
Fix early returns

ChrisBr:
We recently saw an increase of early exits in our builds because of too many failures.

tl;dr: I _believe_ this is caused when too many tests timeout. We would increment the test failures when a test failed and timed out (aka was claimed by another worker). However, if a test timed out and was claimed by another worker we should ignore the result.

## Detailed Explanation

When a test fails, we first try to requeue the test.

https://github.com/Shopify/ci-queue/blob/f610c2e0a3978fc236d5c1145d80df562207dc50/ruby/lib/minitest/queue.rb#L247

However, if the test timed out and picked up and processed by another worker, `requeue` will return false (as there is no need to requeue):

https://github.com/Shopify/ci-queue/blob/bfc2bb1b255280dd7e2a63d50e25d9291980a61e/redis/requeue.lua#L17-L19

In this case, we can also not `acknowledge` the test (as it's already acknowledged):

https://github.com/Shopify/ci-queue/blob/bfc2bb1b255280dd7e2a63d50e25d9291980a61e/redis/acknowledge.lua#L9

At the moment, we would record a failure in this case, however, we need to drop the failure as it's already recorded by another worker.


ruby/lib/minitest/queue.rb
```diff
@@ -249,13 +249,15 @@ def run_from_queue(reporter, *)
...
+        elsif (acknowledged = queue.acknowledge(example)) || !failed
           # If the test was already acknowledged by another worker (we timed out)
           # Then we only record it if it is successful.
           reporter.record(result)
         end
 
-        if !requeued && failed
+        # If the test is failed and not requeued but acknowledged by this worker, then it is a terminal failure
+        # and we have to record it for our circuit breakers.
+        if failed && (!requeued || acknowledged)
```
ChrisBr, line 260: Actually I think this is not correct 🤔

shioyama (approved):

stephsachrajda (approved):

richardnuno (approved):

PR #252, opened 2024-01-15
Safeguard Time.now

ChrisBr:
Safeguard `Time.now`

We had a build which was hanging for a long time. After inspecting the Redis queue, we found that there was one test hanging in the running queue as the expire timestamp was 3 months in the future.

Mocks should be cleaned up when ci-queue runs, however, there might be a case when they leak. This PR makes ci-queue resilient to this.

```
zscore build:unit:018d0bc9-5038-437e-a94e-8976738a051d:running $TEST_ID 
1712077200
```

image

### DIFF
--- a/ruby/lib/ci/queue/circuit_breaker.rb
+++ b/ruby/lib/ci/queue/circuit_breaker.rb
@@ -46,7 +46,7 @@ module CI
         private
 
         def current_timestamp
-          Time.now.to_i
+          CI::Queue.time_now.to_i
         end
       end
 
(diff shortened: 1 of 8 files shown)


ruby/lib/ci/queue/circuit_breaker.rb
```diff
@@ -46,7 +46,7 @@ def message
         private
 
         def current_timestamp
-          Time.now.to_i
+          CI::Queue::Time.now.to_i
```
casperisfine, line 49:
```suggestion
          CI::Queue.time_now.to_i
```

Would be a nicer interface IMO.

PR #253, opened 2024-01-23
Add heartbeat

ChrisBr:
At the moment, the only way to detect if a worker died is when a test has not been processed in the specified timeout. However, for some tests we have to set a rather large timeout so it would take a while to detect a dead worker.

This PR introduces a heartbeat process. When a worker picks up a test we set a started_at timestamp in Redis. Previously, all tests which were started before `Time.now - timeout` were considered lost and could be picked up by another worker. With the introduced heartbeat process we can now consider a test lost when the last heartbeat was `Time.now - max_missed_heartbeats` which is much faster for tests with a large timeout.

This is implemented via 

* a thread which pushes the current test to the heartbeat monitor
* a heartbeat monitor process which runs as a fork and periodically updates the timestamp of the current test

We've implemented this as a sidecar process as some of our tests do not allow e.g. connections to Redis.

### DIFF
--- a/ruby/lib/ci/queue/redis/base.rb
+++ b/ruby/lib/ci/queue/redis/base.rb
@@ -50,6 +50,37 @@ module CI
           [0, 0, 0.1, 0.5, 1, 3, 5]
         end
 
+        def with_heartbeat(id)
+          if heartbeat_enabled?
+            ensure_heartbeat_thread_alive!
+            heartbeat_state.set(:tick, id)
+          end
+
+          yield
+        ensure
+          heartbeat_state.set(:reset) if heartbeat_enabled?
+        end
+
+        def ensure_heartbeat_thread_alive!
+          return unless heartbeat_enabled?
+          return if @heartbeat_thread&.alive?
+
+          @heartbeat_thread = Thread.start { heartbeat }
+        end
+
+        def boot_heartbeat_process!
+          return unless heartbeat_enabled?
+
+          heartbeat_process.boot!
+        end
+
+        def stop_heartbeat!
+          return unless heartbeat_enabled?
+
+          heartbeat_state.set(:stop)
+          heartbeat_process.shutdown!
+        end
+
         def custom_config
           return unless config.debug_log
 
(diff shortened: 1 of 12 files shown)


ruby/lib/minitest/queue.rb
```diff
@@ -256,6 +259,25 @@ def run_from_queue(reporter, *)
...
+    end
+
+    def with_heartbeat(example)
+      finished = false
+      t = Thread.start do
+        heartbeats = queue.config.timeout.to_f * 1_000 # convert to milliseconds
+        while heartbeats > 0 && !finished
+          queue.monitor.tick!(example.id)
+          heartbeats -= 1
+          sleep 0.001
```
casperisfine, line 273: I suspect you did this for testing, but ticking every 1ms is a lot, IMO you can tick every 1 or 2 seconds just fine.

ruby/lib/minitest/queue.rb
```diff
@@ -256,6 +259,25 @@ def run_from_queue(reporter, *)
           reporter.record(result)
         end
       end
+
+      queue.monitor.shutdown!
+    end
+
+    def with_heartbeat(example)
+      finished = false
+      t = Thread.start do
```
casperisfine, line 268:
```suggestion
      t = Thread.start do
        Thread.current.name = "CI::Queue#heartbeat"
```

Not strictly necessary, but it's nice to name threads.

ruby/lib/ci/queue/static.rb
```diff
@@ -48,6 +48,18 @@ def populate(tests, random: nil)
...
       end
 
+      def timeout
+        config.timeout
+      end
+
+      Monitor = Struct.new(:boot!, :shutdown!) do
+        def tick!(_)
+        end
+      end
```
casperisfine, line 58: Lol at `:boot!` struct member. Maybe do a simple PORO ? :)

ruby/lib/ci/queue/redis/base.rb
```diff
@@ -44,6 +44,55 @@ def initialize(redis_url, config)
...
+            child_read.close
+
+            # Check the process is alive.
+            ready_pipe.wait_readable(2)
+            Process.kill(0, @pid)
+            @pipe
+          end
+
+          def shutdown!
+            @pipe.close
```
casperisfine, line 77: We  should wait for the child here, with `Process.wait(@pid)`.

It's not strictly necessary, because `pid` would do it for us, but it's much cleaner.

PR #258, opened 2024-02-06
Fix expiry of static queue

ChrisBr:
Fix expiry of static queue.

### DIFF
--- a/ruby/lib/ci/queue/static.rb
+++ b/ruby/lib/ci/queue/static.rb
@@ -63,7 +63,7 @@ module CI
       end
 
       def expired?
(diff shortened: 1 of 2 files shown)


ruby/lib/ci/queue/static.rb
```diff
@@ -63,7 +63,7 @@ def created_at=(timestamp)
       end
 
       def expired?
-        (@created_at.to_f TEN_MINUTES) < CI::Queue.time_now.to_f
+        (@created_at.to_f + TEN_MINUTES) < CI::Queue.time_now.to_f
```
ChrisBr, line 66: To be honest, we could also just return `false` here, static queue should never be expired.

casperisfine (approved):

PR #267, opened 2024-03-27
Centralize Ruby Version to `.ruby-version`

george-ma:
The `.ruby-version` file is the ecosystem standard for defining a Ruby version.  This PR adds the `.ruby-version` file, ensures a `required_ruby_version` is set, and removes all other references to Ruby in this repository, aligning it with the standard.
> [!IMPORTANT]
> Please verify the following before merging:

Verify that the changes in the PR meets the following requirements or adjust manually to make it compliant:
  - [ ] `.ruby-version` file is present with the correct Ruby version defined
  - [ ] A `required_ruby_version` in your gemspec is set
  - [ ] There is no Ruby version present in the  `dev.yml` Ruby task (before: `- ruby: x.x.x`, after: `- ruby`)
  - [ ] There is no Ruby version/requirement referenced in the `Gemfile` (no lines with `ruby  <some-version>`)
  - [ ] A `Gemfile.lock` is built with the defined Ruby version
  - [ ] The version of Rubocop installed is 1.61.0 or greater
  - [ ] There is no `TargetRubyVersion` defined  in `rubocop.yml` (reads from  `required_ruby_version` on Rubocop 1.61.0)
  - [ ] There is no Ruby argument  present in  `ruby/setup-ruby` Github Actions that do **not**  run on a Ruby matrix (no lines with `ruby-version: “x.x”`)

To establish consistency, the `required_ruby_version`is set to 3.0. If you think that another version is a better fit, please make the applicable changes.

Please merge this PR if it looks good, this PR will be merged if there isn't any activity after 4 weeks.


.github/workflows/tests.yml
```diff
@@ -16,7 +16,6 @@ jobs:
       fail-fast: false
       matrix:
         ruby:
-          - '2.7'
```
ChrisBr, line 19: Usually we keep Ruby versions until they cause some breaking changes. Any reason we're dropping this?

george-ma: Ruby 2.7 reached EOL support last year (Mar 2023) so we've began removing it along with this work. Also open to keeping it if the norm for these repos is to keep until they cause some breaking changes.

ChrisBr: It's open source so we don't really know what the downstream users use.

cc @casperisfine any opinion?

george-ma: 👍🏻 Kept the Ruby 2.7 CI and changed the minimum required Ruby version to be `> 2.7`

ChrisBr (approved):
Thanks

PR #273, opened 2024-04-23
Initialize bisect queue after loading tests

zarifmahfuz:
Follow-up on https://github.com/Shopify/ci-queue/pull/272. We're going clean the test order file given to bisect from the client side instead. Specifically from the file which loads all the tests. We're going to introspect the currently existing tests based on what's available in `Minitest::Test.runnables`. To make sure that ci-queue reads the cleaned test order file, we need to initialize the bisect queue after it runs `load_tests`.


ChrisBr (approved):
Should be fine to flip the order slightly.

PR #276, opened 2024-05-09
Exit with special code upon unexpected error

zarifmahfuz:
We set up ci-queue workers to soft fail on every exit status. We do this mainly to ignore infrastructure failures. Currently, when a test fails due a problem with the test framework, it has an exit status 1 and is ignored. We want to have a unique exit status for such failures to stop ignoring them as soft failure.

### DIFF
--- a/ruby/lib/minitest/queue.rb
+++ b/ruby/lib/minitest/queue.rb
@@ -266,10 +266,10 @@ module Minitest
       reopen_previous_step
       puts red("The heartbeat process died. This worker is exiting early.")
       exit!(41)
-    rescue StandardError => e
+    rescue => error
       reopen_previous_step
-      puts red("This worker died because of a queue or framework error:")
-      puts red("#{e.class}: #{e.message}")
+      puts red("This worker exited because of an uncaught application error:")
+      puts red("#{error.class}: #{error.message}")
       exit!(42)
     end
   end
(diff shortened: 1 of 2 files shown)


shelby-moore-sh (approved):

ChrisBr:
Can you please provide some tophats before we merge it

ruby/lib/minitest/queue.rb
```diff
@@ -266,6 +266,11 @@ def run_from_queue(reporter, *)
       reopen_previous_step
       puts red("The heartbeat process died. This worker is exiting early.")
       exit!(41)
+    rescue StandardError => e
+      reopen_previous_step
+      puts red("This worker died because of a queue or framework error:")
+      puts red("#{e.class}: #{e.message}")
+      exit!(42)
```
casperisfine, line 273: I think the point is to distinguish between "transcient queue errors" (e.g. can't reach the Redis queue), and application errors (e.g. there is a bug in a test that somehow escape Minitest).

As such the message is weird to me. Should be something like `This worker exited because of an uncaught application error.`

ChrisBr:
^^ Reason for this is that I would like to experiment with setting the commit status only on the summary step, this wouldn't work anymore if we don't ignore the exit status of a worker (we can still return a different exit code though). If we put the error into Redis we can still ignore all workers and succeed if all tests passed and no worker crashed.

PR #284, opened 2024-07-24
Report worker crashes to summary command

ChrisBr:
Report worker crashes to summary command.

### DIFF
--- a/ruby/lib/ci/queue/redis/build_record.rb
+++ b/ruby/lib/ci/queue/redis/build_record.rb
@@ -17,6 +17,21 @@ module CI
           @queue.exhausted?
         end
 
+        def report_worker_error(error)
+          redis.pipelined do |pipeline|
+            pipeline.hset(key('worker-errors'), config.worker_id, error.message)
+            pipeline.expire(key('worker-errors'), config.redis_ttl)
+          end
+        end
+
+        def worker_errors
+          redis.hgetall(key('worker-errors'))
+        end
+
+        def reset_worker_error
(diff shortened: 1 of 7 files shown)


ruby/lib/ci/queue/redis/build_record.rb
```diff
@@ -17,6 +17,17 @@ def queue_exhausted?
           @queue.exhausted?
         end
 
+        def report_worker_error(error)
+          redis.pipelined do |pipeline|
+            pipeline.hset(key('worker-errors'), config.worker_id, error.message)
```
casperisfine, line 22: Should we clear this on retry?

ChrisBr: Yeah good point.

ChrisBr: Updated

ruby/lib/minitest/queue.rb
```diff
@@ -268,8 +268,10 @@ def run_from_queue(reporter, *)
       exit!(41)
     rescue => error
       reopen_previous_step
+      queue.report_worker_error(error)
       puts red("This worker exited because of an uncaught application error:")
       puts red("#{error.class}: #{error.message}")
+      puts red("#{error.backtrace}")
```
casperisfine, line 274: That will call `Array#to_s`, that's not what you want:

```
>> puts "#{["foo", "bar"]}"
["foo", "bar"]
```

```suggestion
      error.backtrace.each do |frame|
        puts red(frame)
      end
```

casperisfine (approved):

Issue #292, opened 2024-11-11
Using a Redis server that requires SSL connection with self signed certificate exits 0 with 0 tests run


schneems:
The issue is described in my debugging notes here https://github.com/heroku/heroku-buildpack-ruby/issues/1505#issuecomment-2469218290

## Context

Heroku rolled out a SSL connections by default https://devcenter.heroku.com/changelog-items/2992. However all certificates are self-signed. (as the servers do not "own" the domain `compute-1.amazonaws.com`). Therefore we recommend people disable peer verification https://devcenter.heroku.com/articles/connecting-heroku-redis#connecting-in-ruby.

When this change happened, my CI was reporting that my runs were successful however I never noticed that they were executing 0 tests:

```
-----> Running test command `bundle exec rspec-queue --max-requeues=3 --timeout 180 --queue $REDIS_URL --format documentation || { cat log/test_order.log;  $(exit 1); }`...
Finished in 9.72 seconds (files took 20.02 seconds to load)
0 examples, 0 failures
-----> test command `bundle exec rspec-queue --max-requeues=3 --timeout 180 --queue $REDIS_URL --format documentation || { cat log/test_order.log;  $(exit 1); }` completed successfully
```

While debugging, I tried connecting to a legacy non-ssl url (REDIS_TEMPORARY_URL) and when I made that change everything worked fine.

## Expected

When I try to connect to a redis instance and the connection fails for some reason that it will exit non-zero.

## Actual

It reports 0 tests were executed and then does not fail

## Suggestion

Ideally if the runner cannot connect to redis (for any reason) it would hard fail and ideally give more connection information. I understand that rspec-queue is marked deprecated in the readme, however I'm guessing that this behavior might also persist in other shared connection code-paths.

Beyond improving the connection failure message, if an flag or env var could be added to set the verification mode to none for use with hosted/self-signed certs, that would be helpful as I have no other way of working around this problem at this time.

casperisfine:
> What we could probably do to is to make this opt-in with some sort of `--soft-failure-code 0` or something

@ChrisBr any opinion?

ChrisBr:
>  In my case such a flag would meet my needs for increasing confidence. I'm running a small cluster, 16 nodes, and I don't expect any of them to fail or be unavailable.

[...]

PR #293, opened 2024-11-12
Allow using hosted Redis instances by default

schneems:
Hosted Redis servers use self signed certificates (because they do not own the domain they're running on such as `compute-1.amazonaws.com`) therefore a full SSL connection verification will fail. The fix for that behavior is to disable verification so a self signed certificate can be used [docs from a hosted Redis/Key-value store provider](https://devcenter.heroku.com/articles/connecting-heroku-redis#connecting-in-ruby).

This PR makes the default behavior to allow connecting to self signed Redis servers, and introduces a new flag `--strict-ssl` which will require that the certificate is not self-signed. 

This failing SSL connection behavior was originally reported in https://github.com/Shopify/ci-queue/issues/292, however that issue focues on raising visibility of such failures.

### DIFF
--- a/ruby/lib/ci/queue/configuration.rb
+++ b/ruby/lib/ci/queue/configuration.rb
@@ -3,7 +3,7 @@ module CI
   module Queue
     class Configuration
       attr_accessor :timeout, :worker_id, :max_requeues, :grind_count, :failure_file, :export_flaky_tests_file
-      attr_accessor :requeue_tolerance, :namespace, :failing_test, :statsd_endpoint
+      attr_accessor :requeue_tolerance, :namespace, :failing_test, :statsd_endpoint, :strict_ssl
       attr_accessor :max_test_duration, :max_test_duration_percentile, :track_test_duration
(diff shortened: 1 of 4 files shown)


casperisfine:
Sorry, but you'll have to sign the CLA...

As for the PR, I'm not even sure we need a flag. I can't really think of a security issue caused by CI queue being pointed at the wrong Redis, unless somehow the list of test name is super secret.

schneems:
I have signed the CLA!

schneems:
> I'm not even sure we need a flag.

I don’t feel strongly about it. I generally find changes to defaults easier to get merged if there’s some sort of opt out so I did what I thought would be the path of least resistance. I have no strong reason to force full SSL checking.

It sounds like you prefer fewer config options over greater configurability. Seems reasonable. 

Please confirm and I can rip that out (or feel free to take the otherwise quite simple change and run with it if it’s faster/easier for you).

ChrisBr:
Superseded by https://github.com/Shopify/ci-queue/pull/294

PR #295, opened 2024-11-25
Stop passing unused block

XrXr:
Found with the new warning in Ruby 3.4.


XrXr:
(CI failure due to Python syntax error is unrelated)

PR #296, opened 2024-11-27
Load tests if queue is not initialized

ChrisBr:
Otherwise the worker may exit early with

`0 tests left and 0 workers running.`


ChrisBr:
Python tests are broken on main.

PR #302, opened 2025-01-16
feat: add support for --export-skipped-tests-file

andheiberg:
I've observed failing tests being skipped in a codebase even though the test file wasn't employing the clear and self documenting "skip" method to achieve this. At scale this makes it hard to reason about which tests are actually run and which are skipped.

I wish to create a central file `test_todo.yml` containing all tests that are allowed to be skipped like so:

```
skipped:
  - test/unit/a_test.rb
```

Then I can write a CI job that complains if tests not explicitly marked as skipped centrally is skipped.

### DIFF
--- a/ruby/lib/ci/queue/build_record.rb
+++ b/ruby/lib/ci/queue/build_record.rb
@@ -23,7 +23,7 @@ module CI
         record_stats(stats)
       end
 
-      def record_success(id, stats: nil, skip_flaky_record: false)
+      def record_success(id, stats: nil, skipped: false, requeued: false)
         error_reports.delete(id)
         record_stats(stats)
       end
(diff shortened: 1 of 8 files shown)


ChrisBr:
I would like to avoid storing more data into Redis and adding more options / complexity unless there is a very good reason. 

For skipped tests I'm not sure we need to store it in a distributed way. Maybe we could just do it in a reporter.

```ruby
class SkipAllowListReporter < Minitest::Reporters::BaseReporter
  def initialize(options = {})
    super
    @allow_list = options.delete(:allow_list)
  end


  def record(test)
    super
    raise "boom" if test.skipped? && !allow_list.include?(test.name)
  end
end
```

PR #305, opened 2025-02-06
Fix race condition in recording results

ChrisBr:
I believe there is a race condition between `acknowledge` and `record`.

When we call `acknowledge` it will immediately remove the test from the running set. We only record failure in the next step in `record`. However, we check the running set in `exhausted?` in the summary to decide when the queue is empty. There is a possibility of removing a test from the running set and the summary exiting before we've actually recorded the failure. `acknowledge` and `record` should be a single operation to avoid this.

https://github.com/Shopify/ci-queue/blob/dd85b49edc07fc3ce9c3a1f0399cde5166c04f71/ruby/lib/minitest/queue.rb#L254-L255

https://github.com/Shopify/ci-queue/blob/dd85b49edc07fc3ce9c3a1f0399cde5166c04f71/ruby/lib/ci/queue/redis/base.rb#L104-L106

https://github.com/Shopify/ci-queue/blob/dd85b49edc07fc3ce9c3a1f0399cde5166c04f71/ruby/lib/ci/queue/redis/base.rb#L121-L126

https://github.com/Shopify/ci-queue/blob/dd85b49edc07fc3ce9c3a1f0399cde5166c04f71/ruby/lib/ci/queue/redis/supervisor.rb#L29

https://github.com/Shopify/ci-queue/blob/dd85b49edc07fc3ce9c3a1f0399cde5166c04f71/redis/acknowledge.lua#L7-L8

### DIFF
--- a/ruby/lib/ci/queue/redis/worker.rb
+++ b/ruby/lib/ci/queue/redis/worker.rb
@@ -11,6 +11,157 @@ module CI
       self.requeue_offset = 42
 
       class Worker < Base
+        class ErrorReport
+          class << self
+            attr_accessor :coder
+
+            def load(payload)
+              new(coder.load(payload))
+            end
+          end
+
+          self.coder = Marshal
+
+          begin
+            require 'snappy'
+            require 'msgpack'
+            require 'stringio'
+
+            module SnappyPack
+              extend self
+
+              MSGPACK = MessagePack::Factory.new
+              MSGPACK.register_type(0x00, Symbol)
+
+              def load(payload)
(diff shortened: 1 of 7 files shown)


ruby/lib/ci/queue/redis/worker.rb
```diff
@@ -11,6 +11,157 @@ class << self
       self.requeue_offset = 42
 
       class Worker < Base
+        class ErrorReport
```
ChrisBr, line 14: TODO: Find new location for `ErrorReport` and `FailureFormatter`. This may break public API?

PR #306, opened 2025-02-25
Add timestamp to junit.xml export

ChrisBr:
Add timestamp to junit.xml export.


ruby/lib/minitest/queue/junit_reporter.rb
```diff
@@ -71,6 +71,7 @@ def add_tests_to(testsuites, suite, tests)
             'classname' => suite,
             'assertions' => test.assertions,
             'time' => test.time,
+            'timestamp' => test.start_timestamp,
```
zarifmahfuz, line 74: Where is this `start_timestamp` method defined?

zarifmahfuz (approved):

PR #312, opened 2025-03-12
Print remaining tests

ChrisBr:
Print remaining tests

### DIFF
--- a/ruby/lib/ci/queue/redis/supervisor.rb
+++ b/ruby/lib/ci/queue/redis/supervisor.rb
@@ -39,7 +39,19 @@ module CI
             yield if block_given?
           end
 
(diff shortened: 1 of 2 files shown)


richardnuno (approved):

ruby/lib/ci/queue/redis/supervisor.rb
```diff
@@ -39,7 +39,19 @@ def wait_for_workers
             yield if block_given?
           end
 
-          puts "Aborting, timed out." if time_left <= 0 && !exhausted?
+          if time_left <= 0 && !exhausted?
+            puts "Aborting, timed out."
```
zarifmahfuz, line 43: Something like `Aborting, timed out waiting for the following tests to finish executing:` would make it a bit clearer.

PR #323, opened 2025-04-09
Refactor test ACK to reduce the race condition window

casperisfine:
Because the `ack` and the error registration are so far apart, it's possible that the reporter exit successfully even though one test failed.

The worker first ack that the test was ran, then report the error on stdout and other places, then finally record the error in redis.

This leave a relatively big window during which the reporter might see that the queue is empty, but no errors is reported yet.

To fix that we must ack and report the error atomically, which require to move the ack inside the error reporter.

This commit doesn't fully close the race condition but make it much smaller, and make it much easier to bundle the ack and error registration in a single atomic lua script.

### DIFF
--- a/ruby/lib/ci/queue/redis/build_record.rb
+++ b/ruby/lib/ci/queue/redis/build_record.rb
@@ -56,20 +56,29 @@ module CI
           redis.rpush(key('warnings'), Marshal.dump([type, attributes]))
         end
 
+        Test = Struct.new(:id) # Hack
+
         def record_error(id, payload, stats: nil)
-          redis.pipelined do |pipeline|
-            pipeline.hset(
-              key('error-reports'),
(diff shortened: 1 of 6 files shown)


ruby/lib/ci/queue/redis/build_record.rb
```diff
@@ -56,20 +56,29 @@ def record_warning(type, attributes)
           redis.rpush(key('warnings'), Marshal.dump([type, attributes]))
         end
 
+        Test = Struct.new(:id) # Hack
```
casperisfine, line 59: We should change `#acknowledge` to take the raw ID string.

ChrisBr: I will follow up on that.

Issue #215, opened 2023-05-30
0.28.0 version is not compatible with redis 4.8.1+


hyhuang1218:
This [PR](https://github.com/Shopify/ci-queue/commit/96a754a972db5478cfaefc87946293d366d725bc) introduced the usage of a method `sadd` that always returns an integer value.

See failed CI run https://buildkite.com/shopify/shop-server/builds/167914#01886d13-625e-4fbe-a514-5c289ee654dd

PR #325, opened 2025-05-13
Make leader election idempotent

ChrisBr:
`setnx` which we use during leader election is not safe to retry via Redis `reconnect_attempts` as the command could succeed on the Redis server but we run into a client read timeout. In this case, any subsequent `setnx` calls will return `false` and leader election will fail.

Instead we're now setting a uniq value for `setup` and compare it when we fail.

### DIFF
--- a/ruby/lib/ci/queue/redis/worker.rb
+++ b/ruby/lib/ci/queue/redis/worker.rb
@@ -212,12 +212,10 @@ module CI
                   begin
                     redis.multi do |transaction|
                       transaction.lpush(key('queue'), tests) unless tests.empty?
-                      transaction.set(key('total'), @total)
-                      transaction.set(key('master-status'), 'ready')
+                      transaction.set(key('total'), @total, ex: config.redis_ttl)
+                      transaction.set(key('master-status'), 'ready', ex: config.redis_ttl)
 
                       transaction.expire(key('queue'), config.redis_ttl)
-                      transaction.expire(key('total'), config.redis_ttl)
-                      transaction.expire(key('master-status'), config.redis_ttl)
                     end
                   rescue ::Redis::BaseError, RedisClient::Error => error
                     if !queue_initialized? && attempts < 3


shelby-moore-sh (approved):

ruby/lib/ci/queue/redis/worker.rb
```diff
@@ -201,43 +201,68 @@ def try_to_reserve_lost_test
...
+                duration = measure do
+                  begin
+                    redis.multi do |transaction|
+                      transaction.lpush(key('queue'), tests) unless tests.empty?
+                      transaction.set(key('total'), @total)
+                      transaction.set(key('master-status'), 'ready')
+
+                      transaction.expire(key('queue'), config.redis_ttl)
+                      transaction.expire(key('total'), config.redis_ttl)
+                      transaction.expire(key('master-status'), config.redis_ttl)
```
casperisfine, line 220: The `set` command accept an `ex:` argument since Redis 2.6.12, so we can probably merge these commands with their respective `set`.

ruby/lib/ci/queue/redis/worker.rb
```diff
@@ -201,43 +201,68 @@ def try_to_reserve_lost_test
...
+          value = key('setup')
+
+          begin
+            if master = redis.setnx(key('master-status'), value)
+              yield
+            end
+          rescue ::Redis::BaseError => error
+            puts "Error during leader election: #{error}"
+            if redis.get(key('master-status')) == value
+              master = true
```
casperisfine, line 253: Aren't we supposed to yield here?

ChrisBr: Good catch

ruby/lib/ci/queue/redis/worker.rb
ChrisBr: We don't seem to require a Redis version though. Are we fine changing this anyway?

e.g. we just had a report of somebody using 4.8

https://github.com/Shopify/ci-queue/issues/215
https://github.com/Shopify/ci-queue/blob/main/ruby/ci-queue.gemspec#L34

ruby/lib/ci/queue/redis/base.rb
```diff
@@ -8,6 +8,7 @@ class Base
         TEN_MINUTES = 60 * 10
         CONNECTION_ERRORS = [
           ::Redis::BaseConnectionError,
+          RedisClient::ConnectionError,
```
ChrisBr, line 11: @casperisfine because we're using `without_reconnect` it calls `ensure_connected` on `RedisClient` and hence it throws a `RedisClient::ConnectionError` and not a `::Redis::BaseConnectionError`.

Are we fine including that one here?

ruby/lib/ci/queue/redis/worker.rb
casperisfine: Also note, I'm talking about redis-server here, not the client library.

ruby/lib/ci/queue/redis/base.rb
casperisfine: That means requiring `redis-rb` 5.x. But `redis-rb` shouldn't raise `RedisClient` errors, this seems like a bug to me.

casperisfine: I suppose you can do:

```ruby
CONNECTION_ERRORS  << RedisClient::ConnectionError if defined?(RedisClient::ConnectionError)
```

casperisfine: https://github.com/redis/redis-rb/pull/1312

ruby/lib/ci/queue/redis/base.rb
```diff
@@ -9,7 +9,10 @@ class Base
         CONNECTION_ERRORS = [
           ::Redis::BaseConnectionError,
           ::SocketError, # https://github.com/redis/redis-rb/pull/631
-        ].freeze
+        ]
+
+        # https://github.com/redis/redis-rb/pull/1312
+        CONNECTION_ERRORS << RedisClient::ConnectionError if defined?(RedisClient::ConnectionError)
```
casperisfine, line 15:
```suggestion
        CONNECTION_ERRORS << RedisClient::ConnectionError if defined?(RedisClient::ConnectionError)
        CONNECTION_ERRORS.freeze
```

PR #330, opened 2025-05-21
Fix race condition

ChrisBr:
There is a small race condition between acknowledge and recording the error in which the reporter may see the queue empty and no error recorded and wrongly thinks it's a success.

Follow up from https://github.com/Shopify/ci-queue/pull/323

### DIFF
--- a/ruby/lib/ci/queue/redis/build_record.rb
+++ b/ruby/lib/ci/queue/redis/build_record.rb
@@ -59,20 +59,10 @@ module CI
         Test = Struct.new(:id) # Hack
 
         def record_error(id, payload, stats: nil)
-          # FIXME: the ack and hset should be atomic
-          # otherwise the reporter may see the empty queue before the error
-          # is appended and wrongly think it's a success.
-          if @queue.acknowledge(id)
-            redis.pipelined do |pipeline|
-              pipeline.hset(
-                key('error-reports'),
-                id,
-                payload,
-              )
-              pipeline.expire(key('error-reports'), config.redis_ttl)
-              record_stats(stats, pipeline: pipeline)
-              @queue.increment_test_failed
-            end
+          redis.pipelined do |pipeline|
+            @queue.acknowledge(id, error: payload, pipeline: pipeline)
+            record_stats(stats, pipeline: pipeline)
+            @queue.increment_test_failed(pipeline: pipeline)
           end
           nil
         end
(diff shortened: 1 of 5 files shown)


ruby/lib/ci/queue/redis/build_record.rb
```diff
@@ -59,20 +59,10 @@ def record_warning(type, attributes)
...
 
         def record_error(id, payload, stats: nil)
-          # FIXME: the ack and hset should be atomic
-          # otherwise the reporter may see the empty queue before the error
-          # is appended and wrongly think it's a success.
-          if @queue.acknowledge(Test.new(id))
-            redis.pipelined do |pipeline|
-              pipeline.hset(
-                key('error-reports'),
-                id.dup.force_encoding(Encoding::BINARY),
```
ChrisBr, line 69: @casperisfine do you remember why we sometimes encode the test id (`id.dup.force_encoding(Encoding::BINARY)`) and other times we don't?

casperisfine: To workaround a bug in redis that has since then been fixed: https://github.com/Shopify/ci-queue/issues/84

ruby/lib/ci/queue/redis/build_record.rb
```diff
@@ -59,20 +59,10 @@ def record_warning(type, attributes)
...
-              )
-              pipeline.expire(key('error-reports'), config.redis_ttl)
-              record_stats(stats, pipeline: pipeline)
-              @queue.increment_test_failed
-            end
+          @queue.acknowledge(id, error: payload)
+          redis.pipelined do |pipeline|
+            record_stats(stats, pipeline: pipeline)
+            @queue.increment_test_failed
           end
```
casperisfine, line 66:
```suggestion
          redis.pipelined do |pipeline|
            @queue.acknowledge(id, error: payload)
            record_stats(stats, pipeline: pipeline)
            @queue.increment_test_failed
          end
```

If the return value don't matter, you can pipeline it, no?

casperisfine (approved):
Code look OK. Not sure about the race being fixed or not, I don't have this code in mind enough anymore.

ChrisBr:
@casperisfine ♻️

ruby/lib/ci/queue/static.rb
```diff
@@ -103,12 +103,12 @@ def exhausted?
         @queue.empty?
       end
 
-      def acknowledge(test)
+      def acknowledge(*)
```
casperisfine, line 106: Actually, this takes keywords, so `...` would be simpler.

shelby-moore-sh (approved):

PR #334, opened 2025-05-23
Remove rate limit on reserved tests

shioyama:
Currently, ci-queue only allows one test to be reserved at a time. We are attempting to implement test parallelism within ci-queue jobs, which requires that the server process reserves more than this.

I've made some changes here to support that change:

- Implement `reserved_test` as a set instead of a single test (renamed to `reserved_tests`)
- Only raise on `acknowledge` and `requeue`, not on `reserve` since there is nothing to check when reserving a test.

I've implemented this for everything, i.e. without some configuration to switch it on and off. This simplifies things a lot, because I don't think ensuring that we're not reserving when a test has already been reserved is actually doing much atm. But maybe I'm missing something.

### DIFF
--- a/ruby/lib/ci/queue/redis/worker.rb
+++ b/ruby/lib/ci/queue/redis/worker.rb
@@ -16,7 +16,7 @@ module CI
         attr_reader :total
 
         def initialize(redis, config)
-          @reserved_test = nil
+          @reserved_tests = Set.new
           @shutdown_required = false
           super(redis, config)
         end
@@ -135,7 +135,7 @@ module CI
             argv: [config.max_requeues, global_max_requeues, test_key, offset],
           ) == 1
 
-          @reserved_test = test_key unless requeued
+          reserved_tests << test_key unless requeued
           requeued
         end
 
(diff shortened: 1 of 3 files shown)


ruby/lib/ci/queue/redis/worker.rb
```diff
@@ -14,7 +14,7 @@ class Worker < Base
         attr_reader :total
 
         def initialize(redis, config)
-          @reserved_test = nil
+          @reserved_tests = Set.new
```
ChrisBr, line 17: I guess it would be difficult to send the process ID back up and scope by that?

shioyama: Hmm, not sure that would be possible 🤔 

We have a master process that is reserving tests, which forked processes pull from. The master process fills the queue as soon as it has less than the max (currently 1).

When we are reserving a test, therefore, we don't know which process will end up running it, so we can't check for a "mismatch" at that point.

However, we might be able to improve the check in `acknowledge` such that when a worker process takes a test (`pop`s it from the server via drb), we track its process ID and associated it with the reserved test. Then in theory when the result comes back, we could check that the reserved test exists, _and_ that the pid we associated with it matches the worker that is trying to record the result.

I'm not sure that we can get the pid though from a request to the drb server. 🤔

shioyama: Ok so we can indeed check that the worker recording a result for a test matches the pid for the worker that asked for that same test. I don't think it would be easy to do that on the gem side of this, though. And I'm not really convinced that having that check would do much for us.

OTOH I don't see how we can do anything about `reserve`, given that when a server process reserves a test, it doesn't know which worker will end up taking it. So there's no check to be made at that moment.

ruby/lib/ci/queue/redis/worker.rb
```diff
@@ -13,9 +13,10 @@ class << self
       class Worker < Base
         attr_reader :total
 
-        def initialize(redis, config)
-          @reserved_test = nil
+        def initialize(redis, config, max_reserved_tests: 1)
```
shioyama, line 16: @ChrisBr I added a parameter we can use to ensure we are not reserving more tests than we allow in the in-memory queue. This is fairly easy and I think would catch the obvious issues? It would be a duplicate check since we already check this in the parallelization server but can't be too careful.

shioyama:
Ok I extracted the stuff not related to reserving tests to https://github.com/Shopify/ci-queue/pull/337, which should be uncontroversial. Then we can deal with the change to reserved tests on its own.

shioyama:
Updated this PR to only include the part about reserving tests; the rest is in #337.

ruby/lib/ci/queue/redis/worker.rb
```diff
@@ -158,25 +158,26 @@ def release!
...
         end
 
         def raise_on_mismatching_test(test)
-          if @reserved_test == test
-            @reserved_test = nil
+          if reserved_tests.include?(test)
+            reserved_tests.delete(test)
           else
-            raise ReservationError, "Acknowledged #{test.inspect} but #{@reserved_test.inspect} was reserved"
+            raise ReservationError, "Acknowledged #{test.inspect} but only #{reserved_tests.map(&:inspect).join(", ")} reserved"
```
casperisfine, line 173:
```ruby
unless reserved_tests.delete?(test)
  raise ...
end

```

casperisfine (approved):

PR #343, opened 2025-06-30
Fix ack

ChrisBr:
`sadd` returns the number of added entries so it always returns a truthy value. We have to `== 1`. 

This can result in race conditions for lost tests when another worker marks a lost test as processed but then fails on the original worker. 

image

### DIFF
--- a/redis/acknowledge.lua
+++ b/redis/acknowledge.lua
@@ -8,7 +8,7 @@ local error = ARGV[2]
 local ttl = ARGV[3]
 redis.call('zrem', zset_key, test)
 redis.call('hdel', owners_key, test)  -- Doesn't matter if it was reclaimed by another workers
-local acknowledged = redis.call('sadd', processed_key, test)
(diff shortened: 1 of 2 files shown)


richardnuno (approved):

redis/acknowledge.lua
```diff
@@ -8,7 +8,7 @@ local error = ARGV[2]
 local ttl = ARGV[3]
 redis.call('zrem', zset_key, test)
 redis.call('hdel', owners_key, test)  -- Doesn't matter if it was reclaimed by another workers
-local acknowledged = redis.call('sadd', processed_key, test)
+local acknowledged = redis.call('sadd', processed_key, test) == 1
```
casperisfine, line 11: I think the mistake was to think Lua consider `0` as falsey.

casperisfine (approved):

redis/acknowledge.lua
ChrisBr: That's right 🤦

PR #348, opened 2025-07-01
Fix reporter

ChrisBr:
We have to report a non 0 exit status worker died as well.

### DIFF
--- a/ruby/lib/minitest/queue/build_status_reporter.rb
+++ b/ruby/lib/minitest/queue/build_status_reporter.rb
@@ -129,7 +129,7 @@ module Minitest
 
         puts aggregates
 
-        if supervisor.time_left.to_i <= 0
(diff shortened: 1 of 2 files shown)


ruby/lib/minitest/queue/build_status_reporter.rb
```diff
@@ -167,7 +167,11 @@ def report
 
       def success?
         build.error_reports.empty? &&
-          build.worker_errors.empty?
+          build.worker_errors.empty? &&
+          build.queue_exhausted? &&
+          !supervisor.max_test_failed? &&
+          !all_workers_died? &&
```
ChrisBr, line 173: The only missing case should've been `all_workers_died` because the other cases are implicitly checked through `error_reports`. I've added them nevertheless to be 100% sure though.

PR #349, opened 2025-07-01
Fix race in requeue

ChrisBr:


redis/requeue.lua
```diff
@@ -31,6 +32,8 @@ end
 redis.call('hincrby', requeues_count_key, '___total___', 1)
 redis.call('hincrby', requeues_count_key, test, 1)
 
+redis.call('hdel', error_reports_key, test)
```
casperisfine, line 35: If we're doing `hdel` here, the `transaction.hdel(key('error-reports'), id)` becomes redundant.

But also since they were in a transaction, I don't see what this changes.

ChrisBr: This is requeue which is **not** in a transaction. It's called from the queue.

1) We call requeue from queue
2) We go through the reporters
3) `BuildStatusRecorder` will call `record_success` which will remove the error report

This opens a window for a race because as soon as it's requeued a new worker can run it and report a new error report which would could be deleted by the worker which did the requeue.

It's also not redundant because we now call `record_requeue` (instead of `record_success`) which will only report the status. We still need to do the delete in `record_success` which is in a transaction with `acknowledge` already.

https://github.com/Shopify/ci-queue/blob/d154d85f4356240e20c9fa3d407ecc10b8aeb7e7/ruby/lib/minitest/queue.rb#L189-L191

ChrisBr: If we call `requeue` we don't call `record_success` anymore but `record_requeue` which is only setting the stats.

```ruby
        def record_requeue(id, stats: nil)
          redis.pipelined do |pipeline|
          end
        end
```

ChrisBr: Before we were calling `record_success` on requeues and were 

1) deleting the error
2) setting the stats

As we now delete the error report from inside the `requeue` lua script I've created a dedicated method.

The alternative would be to alter `record_success` to something like this

```
        def record_success(id, stats: nil, skip_flaky_record: false, acknowledge: true)
          _, error_reports_deleted_count, requeued_count, _ = redis.multi do |transaction|
            if acknowledge
              @queue.acknowledge(id, pipeline: transaction)
              transaction.hdel(key('error-reports'), id)
            end
            transaction.hget(key('requeues-count'), id)
            record_stats(stats, pipeline: transaction)
          end
          record_flaky(id) if !skip_flaky_record && (error_reports_deleted_count.to_i > 0 || requeued_count.to_i > 0)
          nil
        end
```

But it's difficult because the return values would be incorrect in that case?

casperisfine: Well. `error_reports_deleted_count ` is now already always `0` because it's deleted inside `acknowledge`.

ChrisBr: Yeah it would be fixed with this change because we don't need the conditional anymore.

casperisfine (approved):

PR #355, opened 2025-08-20
Make the CI Queue reporters work in parallel

gmcgibbon:
Make `Minitest::Queue::OrderReporter` parallel-friendly. This is essentially done by appending and truncating to the log file instead of writing to it. The object also doesn't depend on the start hook anymore to initialize the log file, because that isn't called in forked workers.

### DIFF
--- a/ruby/lib/minitest/queue/order_reporter.rb
+++ b/ruby/lib/minitest/queue/order_reporter.rb
@@ -4,21 +4,29 @@ require 'minitest/reporters'
 class Minitest::Queue::OrderReporter < Minitest::Reporters::BaseReporter
   def initialize(options = {})
     @path = options.delete(:path)
+    @file = nil
     super
   end
 
   def start
-    @file = File.open(@path, 'w+')
     super
+    file.truncate(0)
   end
 
   def before_test(test)
     super
-    @file.puts("#{test.class.name}##{test.name}")
-    @file.flush
+    file.puts("#{test.class.name}##{test.name}")
+    file.flush
   end
 
   def report
-    @file.close
+    file.close
+  end
+
+  private
+
+  def file
+    @file ||= File.open(@path, 'a+')
   end
 end
+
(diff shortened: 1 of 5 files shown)


gmcgibbon:
This passes locally. I assume CI is broken?

ruby/lib/minitest/queue/order_reporter.rb
```diff
@@ -4,21 +4,28 @@
...
   def initialize(options = {})
     @path = options.delete(:path)
+    @file = nil
     super
   end
 
   def start
-    @file = File.open(@path, 'w+')
-    super
+    file.truncate(0)
```
davidstosik, line 12: Is `#start` called only once before forking?

If not, is there a risk that you truncate the few first logs in the case the first tests ran very quickly?

Also, is it okay not to call `super`?

ruby/test/minitest/queue/order_reporter_test.rb
```diff
@@ -7,16 +7,41 @@ class OrderReporterTest < Minitest::Test
...
     def test_before_test
+      @reporter.start
       @reporter.before_test(runnable('a'))
       @reporter.before_test(runnable('b'))
       @reporter.report
       assert_equal ['Minitest::Test#a', 'Minitest::Test#b'], File.readlines(log_path).map(&:chomp)
     end
 
+    unless truffleruby?
+      def test_forking
```
davidstosik, line 27: Do you mean to run this test only when not running the test suite with Truffle Ruby?
Does that mean the feature implemented in this PR is not compatible with Truffle Ruby?
Do we need a fallback? Will ci-queue not work with Truffle Ruby anymore?

davidstosik (approved):
Thank you for working on this!
 
Can we make sure calling `file.truncate` in `#start` does not risk truncating the first few logs?

ruby/lib/minitest/queue/order_reporter.rb
gmcgibbon: Yes it is called once before forking, so the truncate will behave like opening the file in write mode and rewriting sections of it in a non-forked setup.

gmcgibbon: Looks like we call https://github.com/minitest/minitest/blob/7c907429e8e7ec7c3b6ee0c7045e67249f4ad505/lib/minitest.rb#L864

Probably not needed, I'll remove it.

PR #358, opened 2025-08-25
Eagerly open file for appending in `OrderReporter`

shioyama:


davidstosik (approved):
I have the uneasy feeling that opening the file too early might cause other issues, but I can't pinpoint a proper reason.
This might just be fine as-is. 👍🏻 

Oh here's one: are there scenarios in which `OrderReporter` would be initialized (meaning it will open the file), but `#report` wont be called (meaning it wouldn't close it)?

PR #371, opened 2026-01-15
Prevent skipped examples from hanging rspec-queue

eliotsykes:
Fixes issue with skipped examples (i.e. those written with `xit`, `xcontext`, etc.) causing the rspec-queue command to hang.

This may have been introduced in https://github.com/Shopify/ci-queue/pull/323 which looks like it removed acknowledgment of skipped examples unintentionally in this change: https://github.com/Shopify/ci-queue/pull/323/changes#diff-974bf1c9f4ea313432db2adde0f2a16f9bd0203e938d49acea6cdb5e12b69082

### DIFF
--- a/ruby/lib/rspec/queue.rb
+++ b/ruby/lib/rspec/queue.rb
@@ -226,6 +226,7 @@ module RSpec
             dup.mark_as_requeued!(reporter)
             return true
           else
+            reporter.acknowledge if skipped?
             super(reporter)
           end
         else
@@ -422,7 +423,7 @@ module RSpec
       end
 
       def acknowledge
-        @queue.acknowledge(@example)
+        @queue.acknowledge(@example.id)
       end
     end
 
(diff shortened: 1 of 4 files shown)


eliotsykes:
Hi maintainers, just checking in if there's anything I need to do to re-run the CLA check?

eliotsykes:
For anyone else hitting this bug with rspec-queue hanging on skipped/pending specs, one workaround is to remove all use of `xit, xdescribe, xscenario, xcontext, xfeature, xspecify` and `:skip`-metadata in your specs.

You could replace those with a tag such as `:techdebt` which you then configure rspec to exclude from test runs with something like this which typically goes in `spec_helper.rb`:

```ruby
# Use :techdebt instead of xit, xdescribe, xscenario, xcontext, xfeature, xspecify and :skip to 
# workaround rspec-queue hanging bug: https://github.com/Shopify/ci-queue/pull/371
config.filter_run_excluding :techdebt
```

And your specs would change like this:
```diff
-xit 'some test' do ...
+it 'some test', :techdebt do ...
```

PR #393, opened 2026-04-08
Fix summary step to wait for retry workers before reporting

ianks:
Two fixes for the retry summary step.

**The race condition**: previously, `wait_for_workers` would see the queue as exhausted and exit immediately on retry runs -- before the retry worker had a chance to re-run the failed tests and clear `error-reports`. This meant the summary step would report failures that were about to be fixed. `wait_for_workers` now stays open on retry runs (`BUILDKITE_RETRY_COUNT > 0`) until `error-reports` is empty or `inactive_workers_timeout` expires.

**The misleading stats line**: when workers die before flushing their per-test stats, the summary line would show "0 failures, 0 errors" in green while simultaneously printing a `FAILED TESTS SUMMARY` below it. `aggregates` now uses `error-reports` as a floor for the failure count, so the line can never be misleadingly green when failures are known.

### DIFF
--- a/ruby/Gemfile.lock
+++ b/ruby/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    ci-queue (0.89.0)
+    ci-queue (0.90.0)
       logger
 
 GEM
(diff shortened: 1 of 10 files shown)


ianks:
Verified end-to-end on https://buildkite.com/shopify/world-shopify-selective-tests/builds/903009 -- `CiQueueRetrySmokeTest#test_passes_only_on_manual_retry` ran through the retry path and passed with `BUILDKITE_RETRY_TYPE=manual`. Final aggregate: 0 failures, 0 errors across 602k tests.

thadcraft-shopify (approved):

PR #394, opened 2026-04-09
Don't count requeued test failures toward circuit breaker

ChrisBr:
report_failure! was called before the requeue check, so successfully requeued tests incremented the consecutive failure counter. With max_consecutive_failures=N and N+ deterministic failures that are all requeued, the circuit breaker fired prematurely and the worker exited, stranding requeued tests in the queue with no worker to process them.

Now report_failure!/report_success! is only called when a test genuinely fails or passes. Requeued tests are invisible to the circuit breaker — they don't increment or reset the consecutive failure counter.


ChrisBr:
The Python failures were fixed in https://github.com/Shopify/ci-queue/pull/382 so it's expected that they're failing on this PR.

PR #401, opened 2026-04-24
feat: add local queue

andheiberg:
We use ci-queue on the CI.

I would like the option of making my local setup more similar to CI.

When running parallel test workers on a single machine it's faster and simpler to use a file based queue.

### DIFF
--- a/ruby/lib/ci/queue.rb
+++ b/ruby/lib/ci/queue.rb
@@ -17,6 +17,7 @@ require 'ci/queue/bisect'
 require 'ci/queue/queue_entry'
 require 'ci/queue/class_resolver'
 require 'ci/queue/file_loader'
+require 'ci/queue/local'
 
 module CI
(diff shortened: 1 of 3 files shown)


ruby/lib/ci/queue.rb
```diff
@@ -81,6 +82,8 @@ def from_uri(url, config)
         Static
       when 'file', nil
         File
+      when 'local'
+        Local
```
andheiberg, line 86: We should follow the pattern from redis and only require the local queue if it's needed.

ruby/lib/ci/queue.rb
```diff
@@ -81,6 +82,8 @@ def from_uri(url, config)
         Static
       when 'file', nil
         File
```
andheiberg, line 84: Oh have I made something redundant?

PR #402, opened 2026-04-29
Add normalize_test_id/test_id_normalizer= to configuration

bitwise-aiden:
## Summary

Adds a `test_id_normalizer=` setter to `CI::Queue::Configuration` that allows callers to provide a custom normalization function for test IDs. This ensures `flaky?` lookups work correctly even when test ID formats differ between the flaky tests list and runtime test IDs.

Accepts both `UnboundMethod` and `Method` objects — the setter converts `Method` to a `Proc` automatically so singleton methods from other modules work without binding issues.

## Performance

Since `flaky?` is called for every test in CI (potentially hundreds of thousands of times), we benchmarked the dispatch overhead of different implementation strategies.

### Setup

- 100k test IDs in the pool (simulating a large test suite)
- 1k flaky test IDs stored in a `Set`
- 10k random test IDs queried per benchmark iteration
- The normalizer is an identity function (`id -> id`) to isolate pure dispatch overhead from any real normalization work

### Results

| Approach | i/s | Per call | vs baseline |
|----------|-----|----------|-------------|
| no override (baseline) | 16.3M | 62ns | -- |
| `define_singleton_method` (no normalizer) | 12.4M | 81ns | 1.31x slower |
| `define_singleton_method` (UnboundMethod) | 12.5M | 80ns | 1.30x slower |
| `define_singleton_method` (Method.to_proc) | 8.8M | 113ns | 1.84x slower |
| ivar proc | 3.8M | 263ns | 4.26x slower |
| ivar `bind_call` | 3.2M | 317ns | 5.16x slower |

### Analysis

**We chose `define_singleton_method`** because:

1. Setting the normalizer bakes it into the method dispatch table as a real method, so subsequent calls have zero dynamic dispatch overhead.
2. With an `UnboundMethod` normalizer, performance is **identical** to not having one set at all (both ~1.30x vs baseline) -- the only cost is the extra `normalize_test_id` method call itself.
3. With a `Method` normalizer (converted via `to_proc`), there's a small additional overhead (1.84x vs baseline) due to the proc indirection -- but this is still 2-3x faster than the ivar-based alternatives.
4. The ivar-based approaches (`bind_call`, proc) are 4-5x slower because they pay dynamic dispatch costs on every single invocation.

For best performance, callers should prefer passing an `UnboundMethod`. Passing a `Method` is supported for ergonomics with a modest overhead.

## Usage

```ruby
config = CI::Queue::Configuration.new(flaky_tests: ["./test/my_test.rb:ATest#test_foo"])

# Using an UnboundMethod (fastest)
config.test_id_normalizer = Module.new {
  define_method(:normalize) { |id| id.sub(%r{^\./}, "") }
}.instance_method(:normalize)

test = Struct.new(:id).new("test/my_test.rb:ATest#test_foo")
config.flaky?(test) # => true
```

```ruby
# Using a Method (singleton method from a module)
module MyNormalizer
  def self.normalize(id)
    id.sub(%r{^\./}, "")
  end
end

config.test_id_normalizer = MyNormalizer.method(:normalize)
```

Without a normalizer, `flaky?` uses exact string matching (identity function):

```ruby
config = CI::Queue::Configuration.new(flaky_tests: ["ATest#test_foo"])

test = Struct.new(:id).new("ATest#test_foo")
config.flaky?(test) # => true

test = Struct.new(:id).new("./test/my_test.rb:ATest#test_foo")
config.flaky?(test) # => false
```

The normalizer is applied to **both** the stored flaky test IDs and the incoming test ID at lookup time, so mismatches in formatting are resolved regardless of which side has the canonical form:

```ruby
config = CI::Queue::Configuration.new(flaky_tests: ["  ATest#test_foo  "])
config.test_id_normalizer = Module.new {
  define_method(:normalize) { |id| id.strip }
}.instance_method(:normalize)

test = Struct.new(:id).new("ATest#test_foo")
config.flaky?(test) # => true
```

## Changes

- `ruby/lib/ci/queue/configuration.rb` -- Added `test_id_normalizer=`, `normalize_test_id`, and `lazy_normalized_flaky_tests`; updated `flaky?` to normalize both sides
- `ruby/test/ci/queue/configuration_test.rb` -- Added tests for `flaky?` with and without a normalizer, including `Method` input

### DIFF
--- a/ruby/lib/ci/queue/configuration.rb
+++ b/ruby/lib/ci/queue/configuration.rb
@@ -172,7 +172,21 @@ module CI
       end
 
       def flaky?(test)
-        @flaky_tests.include?(test.id)
+        lazy_normalized_flaky_tests.include?(normalize_test_id(test.id))
+      end
+
+      def normalize_test_id(id)
+        id
+      end
+
+      def lazy_normalized_flaky_tests
+        @normalized_flaky_tests ||= @flaky_tests.to_set { |test_id| normalize_test_id(test_id) }
+      end
+
+      def test_id_normalizer=(normalizer)
+        normalizer = normalizer.to_proc if Method === normalizer
+        define_singleton_method(:normalize_test_id, normalizer)
+        @normalized_flaky_tests = nil
       end
 
       def seed
--- a/ruby/Gemfile.lock
+++ b/ruby/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    ci-queue (0.94.0)
+    ci-queue (0.95.0)
       logger
 
 GEM
--- a/ruby/lib/ci/queue/version.rb
+++ b/ruby/lib/ci/queue/version.rb
@@ -2,7 +2,7 @@
 
 module CI
   module Queue
-    VERSION = '0.94.0'
+    VERSION = '0.95.0'
     DEV_SCRIPTS_ROOT = ::File.expand_path('../../../../../redis', __FILE__)
     RELEASE_SCRIPTS_ROOT = ::File.expand_path('../redis', __FILE__)
   end
--- a/ruby/test/ci/queue/configuration_test/normalizer_test.rb
+++ b/ruby/test/ci/queue/configuration_test/normalizer_test.rb
@@ -1,0 +1,89 @@
+module CI
+  module Queue
+    class ConfigurationTest < Minitest::Test
+      class NormalizerTest < Minitest::Test
+        module ExampleHelper
+          extend self
+
+          def normalize(id)
+            id.sub(%r{^\./}, "")
+          end
+
+          def strip(id)
+            id.strip
+          end
+        end
+
+        def test_flaky_without_normalizer
+          config = Configuration.new(flaky_tests: Set["ATest#test_foo", "BTest#test_bar"])
+
+          test = Struct.new(:id).new("ATest#test_foo")
+          assert config.flaky?(test)
+
+          test = Struct.new(:id).new("CTest#test_baz")
+          refute config.flaky?(test)
+        end
+
+        def test_flaky_with_instance_method_normalizer
+          config = Configuration.new(flaky_tests: Set["./test/my_test.rb:ATest#test_foo", "./test/my_test.rb:BTest#test_bar"])
+          config.test_id_normalizer = ExampleHelper.instance_method(:normalize)
+
+          test = Struct.new(:id).new("test/my_test.rb:ATest#test_foo")
+          assert config.flaky?(test)
+
+          test = Struct.new(:id).new("test/my_test.rb:CTest#test_baz")
+          refute config.flaky?(test)
+        end
+
+        def test_flaky_with_method_normalizer
+          config = Configuration.new(flaky_tests: Set["./test/my_test.rb:ATest#test_foo", "./test/my_test.rb:BTest#test_bar"])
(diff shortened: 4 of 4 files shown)


ruby/lib/ci/queue/configuration.rb
```diff
@@ -172,7 +172,21 @@ def max_duration=(duration)
...
-        @flaky_tests.include?(test.id)
+        lazy_normalized_flaky_tests.include?(normalize_test_id(test.id))
+      end
+
+      def normalize_test_id(id)
+        id
+      end
+
+      def lazy_normalized_flaky_tests
+        @normalized_flaky_tests ||= @flaky_tests.map { normalize_test_id(_1) }
```
acaron: Configuration.load_flaky_tests returns a Set, and the old code used Set#include? for O(1) lookup. map converts it to an Array, so lazy_normalized_flaky_tests.include? becomes O(n) for every test. Since flaky? is called for every test, this can be a large regression.

 Suggested fix:                                                                                                                                                    
                                                                                                                                                                   
 ```ruby                                                                                                                                                           
   @normalized_flaky_tests ||= @flaky_tests.map { normalize_test_id(_1) }.to_set                                                                                   
 ```

Or to avoid the array allocation entirely maybe something like:

 ```ruby                                                                                                                                                           
   def lazy_normalized_flaky_tests                                                                                                                                 
     @normalized_flaky_tests ||= @flaky_tests.each_with_object(Set.new) do |test_id, set|                                                                          
       set << normalize_test_id(test_id)                                                                                                                           
     end                                                                                                                                                           
   end                                                                                                                                                             
 ```                                                                                                                                                               
To confirm if this works:                                                                                                                                   
 ```ruby                                                                                                                                                           
   @normalized_flaky_tests ||= @flaky_tests.to_set { |test_id| normalize_test_id(test_id) }                                                                        
 ```

acaron: and test to avoid regression maybe

ruby/test/ci/queue/configuration_test.rb
```diff
@@ -200,6 +200,58 @@ def test_new_lazy_load_test_helpers_env
       assert_equal ["test/test_helper.rb", "test/support/helper.rb"], config.lazy_load_test_helper_paths
     end
 
+    def test_flaky_without_normalizer
+      config = Configuration.new(flaky_tests: ["ATest#test_foo", "BTest#test_bar"])
```
acaron: This mock hides that it is technically a set

ruby/lib/ci/queue/configuration.rb
```diff
@@ -172,7 +172,21 @@ def max_duration=(duration)
...
+      end
+
+      def test_id_normalizer=(normalizer)
+        normalizer = normalizer.to_proc if Method === normalizer
+        define_singleton_method(:normalize_test_id, normalizer)
+        @normalized_flaky_tests = nil
```
acaron, line 189: I recommend adding a test where flaky? is called once before setting the normalizer, then the normalizer is set, then flaky? works with normalized IDs. That proves the lazy cache doesn’t stay stale.

ruby/lib/ci/queue/configuration.rb
bitwise-aiden: Good shout. I had that in a version, but didn't make it all the way through.

acaron (approved):

joshbeckman (approved):

thadcraft-shopify (approved):

PR #405, opened 2026-04-30
Make HeartbeatProcess#tick! lease argument backward-compatible

acaron:
The `lease` parameter was added to `HeartbeatProcess#tick!` in v0.85.0, changing the signature from 1 argument to 2. Callers still using the old 1-argument signature crash with `ArgumentError: wrong number of arguments (given 1, expected 2)` when upgrading.

Make `lease` optional so both call sites work. When omitted, `lease` is excluded from the heartbeat payload entirely, matching the pre-v0.85.0 wire format.

Also cleans up `instance_variable_set` usage in the heartbeat tests by introducing a `PipedHeartbeatProcess` test subclass with proper constructor injection.

co-authored by AI

### DIFF
--- a/ruby/test/ci/queue/redis/heartbeat_process_test.rb
+++ b/ruby/test/ci/queue/redis/heartbeat_process_test.rb
@@ -21,6 +21,18 @@ class CI::Queue::Redis::Base::HeartbeatProcessTest < Minitest::Test
     end
   end
 
+  class PipedHeartbeatProcess < CI::Queue::Redis::Base::HeartbeatProcess
+    attr_reader :pipe
+
+    def initialize(pipe, *args)
+      super(*args)
+      @pipe = pipe
+    end
+
+    def boot!; end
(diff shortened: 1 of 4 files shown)


ruby/test/ci/queue/redis/heartbeat_process_test.rb
```diff
@@ -71,12 +83,12 @@ def test_restart_counter_resets_after_success
   end
 
   def test_tick_does_not_allocate_tick_marker_string
-    @hp.instance_variable_set(:@pipe, FakePipe.new)
```
acaron, line 74: Adding PipedHeartbeatProcess to remove usage of instance_variable_set

ianks (changes requested):
Shouldn't we just fix the callsite? it doesn't make much sense to have a nilable `lease`

acaron:
Closing in favour of fixing callsite by fetching lease

PR #407, opened 2026-04-30
add compatibility for minitest 6

catwomey:
minitest 6.0 shipped [breaking API changes](https://minite.st/docs/History_rdoc.html#600--2025-12-17) that affect ci-queue. This PR adds minitest 6 support while remaining backwards compatible with minitest 5.11+.

   ### Breaking changes in minitest 6

   **1. `Minitest.run_one_method(klass, name)` — removed**

   ci-queue used this to run individual tests in `SingleExample#run` and `LazySingleExample#run`. Replaced with `klass.new(name).run` which works on both minitest 5 and 6.

   **2. `Minitest.__run` — renamed to `Minitest.run_all_suites`**

   ci-queue prepends a module onto `Minitest`'s singleton class to override `__run` with its Redis-backed poll loop. minitest 6 renamed this method, so the override no longer fires. We now define both `__run` and `run_all_suites` on the prepended module — only the one matching the active minitest version is called;
 the other is inert.

   **3. Plugin loading — now opt-in**

   minitest 5 auto-discovered plugins via `load_plugins` inside `Minitest.run`. minitest 6 removed this. Without it, `minitest-reporters`' `plugin_minitest_reporter_init` never fires, so its `DelegateReporter` is never created. This meant ci-queue's reporters (including `BuildStatusRecorder`) were set in
 `Minitest::Reporters.reporters` but never wired into the `CompositeReporter` that `Minitest.run` actually calls `record` on.

   The consequence: `BuildStatusRecorder#record` never ran → `acknowledge` never called → Redis running set never cleared → `exhausted?` never returned true → **the poll loop hung forever**.

   Fixed by calling `Minitest.load_plugins` explicitly when ci-queue sets up its reporters, and in `run_tests_in_fork` (used by bisect).

   **4. `Parallel::Executor` thread pool**

   minitest 6 starts N threads (where N = CPU count) before `run_all_suites` and joins them afterward via `shutdown`. ci-queue bypasses the executor entirely — it has its own Redis poll loop — so those threads sit idle on `queue.pop` and `shutdown` blocks forever. We now drain the executor after `Queue.run`
 completes.

   ### Backwards compatibility

   All changes are guarded by `respond_to?` checks. No version string comparisons. minitest 5.11+ continues to work unchanged.

### DIFF
--- a/ruby/Gemfile.lock
+++ b/ruby/Gemfile.lock
@@ -38,6 +38,7 @@ GEM
     lint_roller (1.1.0)
     logger (1.7.0)
     minitest (5.27.0)
+    minitest-mock (5.27.0)
     minitest-reporters (1.7.1)
       ansi
       builder
@@ -106,6 +107,7 @@ GEM
 
 PLATFORMS
   arm64-darwin-23
+  arm64-darwin-24
   ruby
 
 DEPENDENCIES
@@ -113,7 +115,8 @@ DEPENDENCIES
   benchmark
   bundler
   ci-queue!
-  minitest (~> 5.11)
+  minitest (>= 5.11)
+  minitest-mock
   minitest-reporters (~> 1.1)
   mocha
   msgpack
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,6 +22,13 @@ jobs:
           - '3.4'
           - '4.0'
           - 'truffleruby'
+        minitest:
+          - '~> 5.11'
+          - '~> 6.0'
+        exclude:
+          # minitest 6 requires Ruby >= 3.2
+          - ruby: '3.1'
+            minitest: '~> 6.0'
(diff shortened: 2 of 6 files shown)


ruby/Gemfile.lock
```diff
@@ -106,14 +107,14 @@ GEM
 
 PLATFORMS
   arm64-darwin-23
-  ruby
```
gregdodd, line 109: do we still want ruby here?

gregdodd (approved):

ruby/Gemfile.lock
catwomey: Good callout, bundler only included my current system.  

Ran `bundle lock --add-platform ruby` to add it back in the amended commit fb167a2776b8df2f8742c3472b40fba3006cd4ed.

PR #411, opened 2026-05-21
fix incomplete file

MoumenKhadr:
### TL;DR

JUnit report files are now written atomically using a temporary file and rename.

### What changed?

The `JUnitReporter#report` method now writes the XML output to a temporary file in the same directory as the target report path, then atomically renames it to the final destination. If an error occurs during writing, the temporary file is cleaned up and no partial file is left at the report path.

### How to test?

Three new test cases cover the updated behavior:
- **Atomic write**: Verifies the report file is written successfully and no `.tmp` files are left behind after a normal run.
- **Error handling**: Simulates a mid-write crash and asserts that neither a partial report file nor an orphaned temp file remains.
- **Nested directories**: Confirms that parent directories are still created when they don't exist.

### Why make this change?

Previously, a crash or error mid-write could leave a partial or corrupt XML file at the report path. By writing to a temp file first and renaming only on success, the final report file is either fully written or absent — preventing downstream consumers from reading incomplete JUnit output.

### DIFF
--- a/ruby/lib/minitest/queue/junit_reporter.rb
+++ b/ruby/lib/minitest/queue/junit_reporter.rb
@@ -3,6 +3,7 @@
 require 'minitest/reporters'
 require 'rexml/document'
 require 'fileutils'
+require 'tempfile'
 
 module Minitest
   module Queue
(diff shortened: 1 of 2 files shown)


MoumenKhadr:
* **#411** <a href="https://app.graphite.com/github/pr/Shopify/ci-queue/411?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/ci-queue/411?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

Amr-Aboshama (approved):

PR #419, opened 2026-07-30
Bump ci-queue to v0.97.0

bitwise-aiden:

PR #420, opened 2026-08-06
Stamp per-worker-process execution metadata on test results

bitwise-aiden:
## What

Stamps three per-worker-process execution metadata fields on every minitest result, carried nil-safely through `TestData#to_h` into `log/test_data.json`:

- **`parallel_worker_pid`** — `Process.pid` at result creation.
- **`parallel_worker_test_index`** — 0-based per-process monotonic counter, incremented per execution (requeued runs get their own index). Fork-safe: restarts when the pid changes, so each process incarnation gets a clean `0,1,2,…` sequence.
- **`parallel_worker_id`** — identifier injected by the embedding environment (e.g. a Rails parallel-testing worker number) via `Minitest::Queue.parallel_worker_id=` or `CI_QUEUE_PARALLEL_WORKER_ID`; `nil` when not applicable.

This makes per-worker-process execution order reconstructable downstream:

```sql
PARTITION BY job_id, parallel_worker_id, parallel_worker_pid
ORDER BY parallel_worker_test_index
```

i.e. a SQL query can reproduce any worker's `test_order-w{N}-{pid}.log` from warehouse rows — the prerequisite for warehouse-native test-pollution / requeue-overlap analysis.

## How

- New `Minitest::ParallelWorkerMetadata` accessors module prepended to `Minitest::Result` (fields ride the result object through Marshal/DRb).
- Stamping happens at the top of `Minitest::Queue.handle_test_result` — in the process that ran the test for all in-process flows.
- **First-writer-wins:** embedders that run tests in forked workers and transport results to a central reporting process (e.g. Rails parallelization over DRb, where `handle_test_result` runs server-side) must call `Minitest::Queue.stamp_parallel_worker_metadata(result)` in the worker before sending. Pre-stamped results pass through reporting untouched — otherwise the reporting-side stamp would carry the server's pid and an arrival-order index interleaved across workers.

## Impact on existing consumers

- Fields are additive and nil-safe; `TestDataReporter` is structurally unchanged.
- rspec-queue, junit.xml, `test_order.log`, and Redis build-status/error reports are untouched.
- Downstream Monorail wrappers whitelist payload keys, so the new keys are inert until schema/field lists are updated deliberately.

## Testing

- New unit suite for stamping (setter/env precedence, fork reset, first-writer-wins, nil-safety).
- `TestData` unit tests for stamped/unstamped/accessor-less results.
- Integration assertions in `test_test_data_reporter`: worker id from env, single pid, indexes exactly `0..N-1`, requeued execution ordered before its final run.

Also bumps the version to 0.98.0 for release.

### DIFF
--- a/ruby/Gemfile.lock
+++ b/ruby/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    ci-queue (0.97.0)
+    ci-queue (0.98.0)
       logger
 
 GEM
--- a/ruby/README.md
+++ b/ruby/README.md
@@ -158,6 +158,41 @@ rspec-queue --queue redis://example.com --timeout 600 --report
 
 Because of how `ci-queue` executes the examples, `before(:all)` and `after(:all)` hooks are not supported. `rspec-queue` will explicitly reject them.
 
+### Parallel worker metadata
+
+Each recorded result is stamped, in the process that ran the test, with:
+
+- `parallel_worker_pid`: the pid of the worker process at result creation.
+- `parallel_worker_test_index`: a 0-based, per-process monotonic counter incremented as each test runs (requeued executions get their own index).
+- `parallel_worker_id`: an identifier injected by the embedding environment (e.g. a Rails parallel-testing worker number); `nil` when not applicable.
+
+All three fields are included (nil-safe) in `log/test_data.json` emitted by the test data reporter, making per-worker-process execution order reconstructable downstream (`PARTITION BY parallel_worker_id, parallel_worker_pid ORDER BY parallel_worker_test_index`).
+
+The worker id can be provided either programmatically or via the environment:
+
+```ruby
+Minitest::Queue.parallel_worker_id = 3
+```
+
+Stamping is first-writer-wins. When ci-queue runs tests in-process (the normal
+`minitest-queue` flow), results are stamped automatically as they are recorded.
+Embedding environments that run tests in forked workers and transport results to
+another process (e.g. over DRb to a central reporting server) must stamp in the
+worker **before** sending:
+
+```ruby
+# in the forked worker, after running the test and before the DRb send
+Minitest::Queue.stamp_parallel_worker_metadata(result)
+```
+
+Otherwise the automatic stamp during reporting would capture the reporting
+process's pid and an arrival-order index interleaved across workers. Pre-stamped
+results pass through reporting untouched.
+
+| Variable | Description |
+|---|---|
+| `CI_QUEUE_PARALLEL_WORKER_ID=N` | Sets `parallel_worker_id` for results produced by this process. The setter takes precedence. |
+
 ### Worker circuit breakers
 
 Both runners support two independent, disabled-by-default circuit breakers:
--- a/ruby/README.md
+++ b/ruby/README.md
@@ -138,6 +138,57 @@ The runner also comes with a tool to investigate leaky tests:
 minitest-queue --queue path/to/test_order.log --failing-test 'SomeTest#test_something' bisect -Itest test/**/*_test.rb
 ```
 
+#### Parallel worker metadata
+
+Each result is stamped, as it is recorded in the process that ran the test, with:
+
+- `parallel_worker_pid`: the pid of the process that ran the test.
+- `parallel_worker_test_index`: a 0-based monotonic counter of results recorded by that process (requeued executions get their own index). Restarts at 0 in each forked process.
+- `parallel_worker_id`: an identifier injected by the embedding environment (e.g. a Rails parallel-testing worker number); `nil` when not applicable.
+
+All three fields are included (nil-safe) in `log/test_data.json` emitted by the test data reporter, making per-worker-process execution order reconstructable downstream (`PARTITION BY parallel_worker_id, parallel_worker_pid ORDER BY parallel_worker_test_index`).
+
+The worker id can be provided either programmatically or via the environment:
+
+```ruby
+Minitest::Queue.parallel_worker_id = 3
+```
+
+| Variable | Description |
+|---|---|
+| `CI_QUEUE_PARALLEL_WORKER_ID=N` | Sets `parallel_worker_id` for results produced by this process. Read once per process; the setter takes precedence. |
+
+Stamping is first-writer-wins. When ci-queue runs tests in-process (the normal
+`minitest-queue` flow), results are stamped automatically as they are recorded.
+Embedding environments that run tests in forked workers and transport results to
+another process (e.g. over DRb to a central reporting server) must stamp in the
+worker **before** sending:
+
+```ruby
+# in the forked worker, after running the test and before the DRb send
+Minitest::Queue.stamp_parallel_worker_metadata(result)
+```
+
+Otherwise the automatic stamp during reporting would capture the reporting
+process's pid and an arrival-order index interleaved across workers. Pre-stamped
+results pass through reporting untouched.
+
+Notes:
+
+- `parallel_worker_id` identifies a forked test process *inside* one queue worker.
+  It is unrelated to ci-queue's own `--worker` / `CI::Queue::Configuration#worker_id`,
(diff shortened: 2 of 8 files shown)


FletcherDares (approved):

mdwn:
From River:

Reviewed at `3384eab`. Nice change overall — the fork-reset mirrors the existing `FileLoader#detect_fork!` pattern, the accessors are nil-safe for embedders that don't have them, the new keys are purely additive, and the first-writer-wins contract has real test coverage. A few things I'd want addressed before this lands.

## Blocking-ish

### 1. Thread-based parallelization silently produces wrong data

All three pieces of state — `@parallel_worker_next_test_index`, the pid, and `parallel_worker_id` — live on the `Minitest::Queue` singleton, so they are process-global.

Under thread-based parallelization (Rails `parallelize(workers: N, with: :threads)`) every "worker" shares one pid and one `parallel_worker_id`. The documented partition key:

```sql
PARTITION BY job_id, parallel_worker_id, parallel_worker_pid
ORDER BY parallel_worker_test_index
```

then collapses to a single partition, and the indexes interleave across threads. The result still *looks* like a valid per-worker execution order downstream, which is the worst failure mode for a telemetry field — silently wrong beats obviously missing. `@parallel_worker_next_test_index += 1` also isn't atomic once you're off CRuby's GVL (JRuby/TruffleRuby, where `:threads` mode is most likely to be used).

Two ways out, either is fine:
- Key the counter (and ideally the worker id) on `Thread.current` rather than the module.
- Or state plainly in the README that only fork-based parallelization is supported, and that `parallel_worker_*` must not be trusted under `with: :threads`.

### 2. The version bump doesn't belong in this PR

Commit `3384eab` bumps to 0.98.0. The repo's own [Releasing a New Version](https://github.com/Shopify/ci-queue/blob/main/ruby/README.md#releasing-a-new-version) section says to bump **after** merging changes to `main`, and the history backs that up — #419 was a standalone `Bump ci-queue to v0.97.0`. Bundling it here also guarantees a conflict with any other in-flight PR that does the same. Suggest dropping the commit and doing the release separately.

## Worth fixing

### 3. `ENV` is re-read and re-parsed on every test

`parallel_worker_id` calls `parallel_worker_id_from_env` on every stamp, i.e. once per test execution. With a non-numeric value (`CI_QUEUE_PARALLEL_WORKER_ID=worker-a`, explicitly supported per the tests) that's an `Integer()` raise plus rescue **per test**. This repo has previously shaved per-heartbeat `Symbol#to_s` allocations, so a per-test exception is going to get noticed.

Memoize it, and clear the memo in the same pid-change branch that already resets the counter — that keeps it correct across forks, where a worker id injected post-fork must not be shadowed by the parent's value:

```ruby
def parallel_worker_id
  return @parallel_worker_id if @parallel_worker_id

  @parallel_worker_id_from_env = parallel_worker_id_from_env unless defined?(@parallel_worker_id_from_env)
  @parallel_worker_id_from_env
end
```

### 4. The payload isn't self-sufficient for the query that justifies it

The partition key in the description starts with `job_id`, but `test_data.json` carries neither `job_id` nor ci-queue's own `worker_id` — `TestData#to_h` has no worker or build identity at all. So "a SQL query can reproduce any worker's `test_order-w{N}-{pid}.log`" holds only if the ingestion layer attaches `job_id` out of band. Worth either emitting `queue.config.worker_id` alongside the new fields, or saying explicitly in the README where `job_id` is expected to come from.

### 5. Naming collides with ci-queue's existing `worker_id`

ci-queue already has a first-class `worker_id`: `config.worker_id`, set by `--worker` / inferred from the CI provider, used throughout `CI::Queue::Redis::Worker`, and already emitted in the worker profile payload. That's a *whole queue worker*. The new `parallel_worker_id` is a fork *inside* one of those. Two different "worker" concepts landing in the same telemetry pipeline will get conflated by whoever writes the queries. Not a blocker, but consider something like `intra_worker_*` / `process_worker_*`, or at minimum call out the relationship in the README.

## Nits

- **README placement.** The new `### Parallel worker metadata` section sits right after the deprecated RSpec section, but the feature is minitest-only (rspec-queue is untouched). It belongs under `### Minitest`, and the one-row env table would be better folded into the existing `CI_QUEUE_*` table there rather than starting a second one.
- **README accuracy.** "the pid of the worker process at result creation" — it's actually stamped at the top of `handle_test_result`, after the test has run. Same process in the in-process flow, so the value is right, but the wording matters precisely because the DRb contract hinges on *when* stamping happens.
- **`test_ignores_results_without_accessors` asserts nothing.** The comment says "does not raise", but nothing enforces that intent. `assert_nil Minitest::Queue.stamp_parallel_worker_metadata(plain)` (or `assert_silent`) would make the test fail for the right reason.
- **Unit tests leak global state.** `teardown` restores `@parallel_worker_id` and the env var, but not `@parallel_worker_metadata_pid` / `@parallel_worker_next_test_index` — and `test_index_restarts_when_pid_changes` deliberately mutates the former. Harmless today because every other assertion is relative, but it's a trap for the next test added to this file.
- **`assert_equal (0...failures.size).to_a, ...`** in the integration test — the space before the paren makes Ruby parse it as a grouped expression; worth closing up.

## Note on CI

There are no workflow runs on head SHA `3384eab` — only Graphite's `mergeability_check` reports, and the combined status is `pending`. So the suite hasn't actually been exercised on this branch yet.

---

Review requested by Mike Wilson <mike.wilson@shopify.com> via River.
Slack thread: https://shopify.slack.com/archives/C0BL3BCFL4A/p1786462080803919

bitwise-aiden:
Thanks for the thorough review — addressed in c6d40a8 (and a force-push dropping the version bump). Point by point:

**1. Thread-based parallelization** — Agreed, and it's actually nearer-term than `with: :threads`: DRb servers dispatch each incoming call on its own thread, so an embedder's reporting process could race the counter even on CRuby. Took the fix rather than the disclaimer for atomicity: stamp state is now mutex-guarded (with a thread-safety test), plus a README note that per-worker order reconstruction is only meaningful with **forked** workers — under threads, everything shares one `(worker_id, pid)` partition and the index is record order across threads. I didn't take the `Thread.current` option: thread identity isn't a stable partition key, there's no thread-id field in the schema to carry it, and per-thread counters would produce *colliding* `(pid, index)` pairs within a process.

**2. Version bump** — Fair, that's the documented release flow and the #419 precedent. Dropped from this PR; the bump is parked on a branch and will go up as a standalone PR once this merges.

**3. ENV re-read per test** — Fixed. Memoized keyed on `Process.pid` (rather than clearing in the stamp's pid-change branch) so the getter is fork-safe even when called outside stamping. Documented as "read once per process"; test added.

**4. Payload self-sufficiency** — Documented rather than emitted. `test_data.json` has never carried build/job identity for *any* field — attaching it is the ingestion layer's job (e.g. Core's wrapper merges `buildkite_job_id`), and emitting `queue.config.worker_id` from `TestData` would mean plumbing queue access into a class that deliberately has none, duplicating what the pipeline already does. README now states where `job_id` is expected to come from.

**5. Naming** — Keeping `parallel_worker_*`, but the relationship is now called out in the README ("a forked test process *inside* one queue worker; unrelated to `--worker`/`config.worker_id`"). Renaming isn't really on the table: `shopify_build_test_results/3.3` is already merged with these exact field names, and Rails 8.1's `ActiveSupport::TestCase.parallel_worker_id` uses the same term for the same concept, so this matches the ecosystem the main embedder lives in.

**Nits** — All taken: section moved under Minitest (kept its own small env table — the env tables are per-feature in this README), stamp-timing wording fixed, `stamp_parallel_worker_metadata` now returns the result or `nil` so the skip paths are assertable (`assert_nil` in both the no-accessor and pre-stamped tests), all five stamp-state ivars reset in setup/teardown, paren spacing fixed.

**CI** — the missing runs were the GitHub Actions outage earlier today; the force-push should re-trigger.

---
*AI generated (pi/claude, reviewed by @aiden before posting)*

mdwn (approved):

